PR #1, opened 2021-10-21
feat: added generic properties and bumped version to 0.2.0

christoshadjiaslanis:
I created this PR as I need support for nested hierarchical data in the properties and not just string kv pairs.

I bumped the minor version since this breaks our first experimental API (fields are no longer public).

Hopefully we can get this merged ASAP and I'll push to crates.io because we need this dependency in the Synth project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "posthog-rs"
 license = "MIT"
-version = "0.2.0"
+version = "0.2.1"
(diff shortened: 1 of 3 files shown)


fuziontech:
Hey @christoshadjiaslanis - Not officially, but I'm happy to get this in. Looks good and tests pass. I'll get this in now. 🚢

fuziontech (approved):
looks good and tests ✅

PR #4, opened 2023-01-13
Implement async PostHog client

NathanFlurry:
Depends on #3.

Breaking changes in `posthog_rs::Error` structure.

PR #3, opened 2023-01-13
Allow configuring API endpoint

NathanFlurry:

PR #5, opened 2023-01-13
Fix batch requests to use batch endpoint instead of capturing events in serial, flatten `properties`

NathanFlurry:
Depends on #3 and #4

**Breaking Change**

Previously, posthog-rs write properties to `properties.props`. This PR flattens the properties to the root of `properties`.


oliverb123:
Thanks for the PR! Sorry it slipped through the cracks. Closing this as the batch capture now actually sends the batch as one, and the properties layout has changed sufficiently that the flattening now longer makes sense, I think.

PR #13, opened 2025-01-06
feat: Send lib name and lib version

robbie-c:
Our other SDKs send their lib name and version with events. Add these here too!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@ use serde::Serialize;
 use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use std::time::Duration;
+use semver::Version;
 
 extern crate serde_json;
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -90,10 +91,39 @@ struct InnerEvent {
...
+
+        let version_str = env!("CARGO_PKG_VERSION");
+        properties.props.insert(
+            "$lib_version".into(),
+            serde_json::Value::String(version_str.into()),
+        );
+
+        if let Ok(version) = version_str.parse::<Version>() {
+            properties.props.insert(
+                "$lib_version__major".into(),
```
robbie-c, line 110: the double underscore is intentional, see https://us.posthog.com/project/2/data-management/properties/01924912-0dda-7ac0-bfed-a5d18e8392f3
Screenshot 2025-01-06 at 15 09 19

rafaeelaudibert (approved):

PR #14, opened 2025-01-06
chore: Add basic github actions PR checks

robbie-c:
I don't have the admin powers to set up the POSTHOG_RS_E2E_TEST_API_KEY environment variable, but when that's done this can be merged :)

Obviously there's a lot more we could do here - but I wanted to get the existing test working on CI, and a test for my previous PR

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -192,7 +192,7 @@ pub mod tests {
     #[test]
     fn inner_event_adds_lib_properties_correctly() {
         // Arrange
-        let mut event = Event::new("test_event", "1234");
+        let mut event = Event::new("unit test event", "1234");
         event.insert_prop("key1", "value1").unwrap();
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -173,19 +173,54 @@ impl Event {
...
+        );
+    }
+
+    #[cfg(feature = "e2e-test")]
     #[test]
     fn get_client() {
-        let client = crate::client(env!("POSTHOG_API_KEY"));
+        use std::collections::HashMap;
+
+        let api_key = std::env::var("POSTHOG_RS_E2E_TEST_API_KEY").unwrap();
```
rafaeelaudibert, line 217: Can you add an `.env.example` file to make it clear that we need this to run tests locally? You could also add a section to the README.md explaining that one needs to grab an API key from their instance to be able to run tests

PR #23, opened 2025-02-23
chore: delete redundant async

va-an:
- delete redundant async
- update README.md (supersedes #8)


oliverb123:
Hey @va-an, appreciate the PR, I actually prefer these functions to be async, as I expect them to need to do async work at some point in the future, and don't want the addition of that to be an API change.

va-an:
Makes sense!
Let me revert this commit and let's fix the README in this PR. What do you think?

PR #25, opened 2025-03-26
make all Event fields pub

aviramha:
Hey!
Thanks for the crate. I needed to modify `timestamp` then realized it's probably useful to have all fields as pub.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -9,12 +9,12 @@ use crate::Error;
 
 #[derive(Serialize, Debug, PartialEq, Eq)]
 pub struct Event {
-    event: String,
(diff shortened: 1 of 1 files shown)


oliverb123:
Hey @aviramha 👋. Thanks for the PR, but I think it would probably be better to expose a setter for the timestamp, rather than exposing the fields themselves (just to make backwards maintainability easier - I may change the type of that field to a `DateTime<Utc>` at some point, for example). I've [opened another PR](https://github.com/PostHog/posthog-rs/pull/30) that exposes a timestamp setter, let me know if that looks like it'll do what you need.

PR #32, opened 2025-08-19
chore: drop async keyword restrictions

laststylebender14:


oliverb123:
Hey @laststylebender14 
Thanks for the PR! The async here is intentionally left as part of the interface to avoid breaking changes if the client ends up doing async work on setup in the future. If you need a sync client, I recommend using the sync feature.

PR #33, opened 2025-09-05
Feature: Allow initializing `Client` in a disabled state

niusia-ua:
This PR updates the `Client` implementation by allowing users to initialize the client in a disabled state if no API key is provided, in which it is possible to capture events without sending them to PostHog.
This is useful for development environments or when you need to conditionally disable event tracking (e.g., based on user privacy settings).

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -74,6 +74,8 @@ impl Event {
     }
 
     /// Set the event timestamp, for events that happened in the past.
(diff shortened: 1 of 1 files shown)


niusia-ua:
Don't pay attention to this PR.
I moved commits from the main branch to a feature branch and re-created the PR (see #34).

PR #34, opened 2025-09-06
Feature: Allow initializing Client in a disabled state

niusia-ua:
This PR updates the `Client` implementation by allowing users to initialize the client in a disabled state if no API key is provided, in which it is possible to capture events without sending them to PostHog.
This is useful for development environments or when you need to conditionally disable event tracking (e.g., based on user privacy settings).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ uuid = { version = "1.13.2", features = ["serde", "v7"] }
 [dev-dependencies]
 dotenv = "0.15.0"
 ctor = "0.1.26"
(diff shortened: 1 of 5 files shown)


rafaeelaudibert:
@niusia-ua It's still not our top priority, but we've recently formed a new [Team Client Libraries](https://posthog.com/teams/client-libraries) in hopes of increasing SDK support across the board. This means Rust will get some love *eventually* but I can't confirm it's now.

rafaeelaudibert:
You can already see some love from the internal team in the most recent PRs

image

PR #36, opened 2025-09-18
initial commit

dmarticus:


src/feature_flags.rs
```diff
@@ -0,0 +1,693 @@
...
+) -> Result<FlagValue, InconclusiveMatchError> {
+    if !flag.active {
+        return Ok(FlagValue::Boolean(false));
+    }
+
+    let conditions = &flag.filters.groups;
+
+    // Sort conditions to evaluate variant overrides first
+    let mut sorted_conditions = conditions.clone();
+    sorted_conditions.sort_by_key(|c| if c.variant.is_some() { 0 } else { 1 });
```
haacked: We have to clone all the conditions and then sort them on every call to `match_feature_flag`.

This sort order is deterministic based on flag definitions. It never changes between evaluations. What if we pre-sorted it when loading flag definitions from the cache in `local_evaluation.rs` in the `update` method.

Even better would be to make the flag definitions already pre-sorted so none of these clients have to do it. But that's beyond the scope of this PR. 😄

src/client/blocking.rs
```diff
@@ -51,13 +104,175 @@ impl Client {
...
+
+    /// Evaluate a feature flag locally (requires feature flags to be loaded)
+    pub fn evaluate_feature_flag_locally(
+        &self,
+        flag: &FeatureFlag,
+        distinct_id: &str,
+        person_properties: &HashMap<String, serde_json::Value>,
+    ) -> Result<FlagValue, Error> {
+        match_feature_flag(flag, distinct_id, person_properties)
+            .map_err(|e| Error::Connection(e.message))
```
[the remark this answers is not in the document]
dustinbyrne: This looks valid, seems worth adding `InconclusiveMatchError` to the `Error` enum

src/endpoints.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub fn determine_server_host(host: Option<String>) -> String {
+        let host_or_default = host.unwrap_or_else(|| DEFAULT_HOST.to_string());
+        let trimmed_host = host_or_default.trim_end_matches('/');
+
+        match trimmed_host {
+            "https://app.posthog.com" | "https://us.posthog.com" => {
+                US_INGESTION_ENDPOINT.to_string()
+            }
+            "https://eu.posthog.com" => EU_INGESTION_ENDPOINT.to_string(),
+            _ => host_or_default,
```
dustinbyrne:
```suggestion
            _ => trimmed_host
```

src/client/async_client.rs
```diff
@@ -10,30 +16,87 @@ use super::ClientOptions;
...
 
-        self.client
-            .post(&self.options.api_endpoint)
-            .header(CONTENT_TYPE, "application/json")
-            .body(payload)
+        let mut url = self.options.endpoints().build_url(Endpoint::Capture);
+        if self.options.disable_geoip {
+            let separator = if url.contains('?') { "&" } else { "?" };
+            url.push_str(&format!("{separator}disable_geoip=1"));
+        }
```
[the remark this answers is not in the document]
dustinbyrne: The blocking client doesn't consider this case

tests/test_local_evaluation.rs
```diff
@@ -0,0 +1,243 @@
...
+    // Test local evaluation
+    let mut properties = HashMap::new();
+    properties.insert("email".to_string(), json!("test@company.com"));
+
+    let result = client
+        .get_feature_flag("feature-b", "user-123", None, Some(properties), None)
+        .await;
+
+    assert!(result.is_ok());
+    // The actual result depends on whether the mock was hit and processed
```
dustinbyrne:
```suggestion
    assert!(result.unwrap() == Some(FlagValue::Boolean(true)));
```

src/client/async_client.rs
```diff
@@ -52,14 +119,217 @@ impl Client {
...
+        groups: Option<HashMap<String, String>>,
+        person_properties: Option<HashMap<String, serde_json::Value>>,
+        group_properties: Option<HashMap<String, HashMap<String, serde_json::Value>>>,
+    ) -> Result<Option<FlagValue>, Error> {
+        let key_str = key.into();
+        let distinct_id_str = distinct_id.into();
+
+        // Try local evaluation first if available
+        if let Some(ref evaluator) = self.local_evaluator {
+            let props = person_properties.clone().unwrap_or_default();
```
matheus-vb: Nit: I don't know how big this can get so maybe it won't make such a difference, but we can avoid cloning the whole map here. You can do something like:
```rust
let empty = HashMap::new();
let props = person_properties.as_ref().unwrap_or(&empty);
```
And just pass that reference around.

src/feature_flags.rs
dmarticus: I'll do this in a future change! trying to limit the already massive scope of this PR lmao

PR #105, opened 2026-04-27
feat: add evaluate_flags() API for single-call flag evaluation

dmarticus:
## Summary

Adds `evaluate_flags(distinct_id, options)` — one `/flags?v=2` round-trip returns a `FeatureFlagEvaluations` snapshot. From the snapshot:

- `is_enabled(key)` / `get_flag(key)` read the cached value and fire a deduplicated `$feature_flag_called` event with full metadata (`$feature_flag_id`, `$feature_flag_version`, `$feature_flag_reason`, `$feature_flag_request_id`).
- `get_flag_payload(key)` reads the payload without firing an event.
- `only_accessed()` and `only([keys])` produce filtered clones for narrower capture attribution.
- `Event::with_flags(&snapshot)` attaches `$feature/<key>` and `$active_feature_flags` to a captured event with no second `/flags` call.

Implements [RFC #1020 — Server SDK Feature Flag Evaluations API](https://github.com/PostHog/requests-for-comments-internal/pull/1020). Mirrors [posthog-python#539](https://github.com/PostHog/posthog-python/pull/539) and [posthog-js#3476](https://github.com/PostHog/posthog-js/pull/3476).

## Deprecations

The legacy single-flag methods are now `#[deprecated]` in favor of `evaluate_flags()` — they keep working but emit a compile warning at every call site:

- `Client::get_feature_flag`
- `Client::is_feature_enabled`
- `Client::get_feature_flag_payload`

## Design decisions

- **Capture wiring:** exposed as `event.with_flags(&snapshot)` rather than a new `capture` parameter — keeps `capture(event)` unchanged and reads naturally in Rust.
- **Empty-distinct_id snapshots short-circuit access events** so `$feature_flag_called` never lands with an empty `distinct_id`.
- **Filter clones don't back-propagate access** to the parent snapshot (each clone gets its own copy of the accessed set).
- **`only_accessed()` returns empty when nothing was accessed** (no fallback-to-all, no warning) — pre-access the flags you want attached.
- **`$feature_flag_error` is comma-joined** combining response-level errors (`errors_while_computing_flags`, `quota_limited`) with per-flag errors (`flag_missing`) so consumers can filter by type.
- **Warning suppression:** new `feature_flags_log_warnings` client option silences the `only(...)` misuse warnings.

> Replaces #104, which was closed when the branch was recreated to clear merge conflicts after rebasing onto `main`.

---

### DIFF
--- a/src/client/async_client.rs
+++ b/src/client/async_client.rs
@@ -359,9 +359,15 @@ impl Client {
         Ok(flags_response.normalize())
     }
 
-    /// Get a specific feature flag value for a user
+    /// Get a specific feature flag value for a user.
     #[must_use = "feature flag result should be used"]
     #[instrument(skip_all, level = "debug")]
+    #[deprecated(
+        since = "0.6.0",
+        note = "Use Client::evaluate_flags() to fetch a snapshot, then call .get_flag(key) on it. \
+                The snapshot deduplicates $feature_flag_called events and supports attaching \
+                rich metadata to captured events via Event::with_flags()."
+    )]
     pub async fn get_feature_flag<K: Into<String>, D: Into<String>>(
         &self,
         key: K,
@@ -407,8 +413,15 @@ impl Client {
         Ok(feature_flags.get(&key_str).cloned())
     }
 
-    /// Check if a feature flag is enabled for a user
+    /// Check if a feature flag is enabled for a user.
     #[must_use = "feature flag enabled check result should be used"]
+    #[deprecated(
+        since = "0.6.0",
+        note = "Use Client::evaluate_flags() to fetch a snapshot, then call .is_enabled(key) \
+                on it. The snapshot deduplicates $feature_flag_called events and supports \
+                attaching rich metadata to captured events via Event::with_flags()."
+    )]
+    #[allow(deprecated)] // calls deprecated get_feature_flag internally
     pub async fn is_feature_enabled<K: Into<String>, D: Into<String>>(
         &self,
         key: K,
@@ -433,8 +446,15 @@ impl Client {
         })
     }
 
-    /// Get a feature flag payload for a user
+    /// Get a feature flag payload for a user.
     #[must_use = "feature flag payload result should be used"]
+    #[deprecated(
+        since = "0.6.0",
+        note = "Use Client::evaluate_flags() to fetch a snapshot, then call \
+                .get_flag_payload(key) on it. Reading the payload from a snapshot is \
+                event-free, matching this method's behavior, and avoids the per-call \
+                /flags request."
+    )]
     pub async fn get_feature_flag_payload<K: Into<String>, D: Into<String>>(
         &self,
         key: K,
--- a/src/client/async_client.rs
+++ b/src/client/async_client.rs
@@ -1,4 +1,5 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::sync::{Arc, Mutex, OnceLock};
 use std::time::Duration;
 
 use reqwest::{header::CONTENT_TYPE, Client as HttpClient};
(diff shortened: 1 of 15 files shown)


src/client/async_client.rs
```diff
@@ -349,4 +508,241 @@ impl Client {
...
+                }
+            }
+        }
+
+        let mut request_id: Option<String> = None;
+        let mut errors_while_computing = false;
+        let mut quota_limited = false;
+
+        if !options.only_evaluate_locally {
+            let response = self.fetch_flag_details(&distinct_id, &options).await?;
```
dustinbyrne: We'll lose any successful local evaluation here if this throws.

Above, we do this:
```rs
if distinct_id.is_empty() || self.options.is_disabled() {
    return Ok(FeatureFlagEvaluations::empty(host));
}
```

Maybe the error here is okay, but worth considering always returning a `FeatureFlagEvaluations`?

src/client/async_client.rs
```diff
@@ -349,4 +508,241 @@ impl Client {
...
+                    locally_evaluated_keys.insert(key);
+                }
+            }
+        }
+
+        let mut request_id: Option<String> = None;
+        let mut errors_while_computing = false;
+        let mut quota_limited = false;
+
+        if !options.only_evaluate_locally {
```
dustinbyrne: If local evaluation is able to evaluate everything, we'll still fall through to hitting `/flags` unless `only_evaluate_locally` is true

src/client/async_client.rs
```diff
@@ -349,4 +508,241 @@ impl Client {
...
+
+fn remote_record_from_detail(key: String, detail: FlagDetail) -> EvaluatedFlagRecord {
+    let metadata = detail.metadata;
+    let reason = detail
+        .reason
+        .and_then(|r| r.description.or(Some(r.code)))
+        .filter(|s| !s.is_empty());
+    let id = metadata.as_ref().map(|m| m.id);
+    let version = metadata.as_ref().map(|m| m.version);
+    let payload = metadata.and_then(|m| m.payload);
```
dustinbyrne: We should normalize the payload to a JSON value

src/client/async_client.rs
```diff
@@ -349,4 +508,241 @@ impl Client {
...
+fn local_record(key: String, value: FlagValue) -> EvaluatedFlagRecord {
+    let (enabled, variant) = match value {
+        FlagValue::Boolean(b) => (b, None),
+        FlagValue::String(s) => (true, Some(s)),
+    };
+    EvaluatedFlagRecord {
+        key,
+        enabled,
+        variant,
+        payload: None,
```
dustinbyrne, line 753: We're skipping payloads for locally evaluated flags here

dustinbyrne: Looks like this is a current gap in the SDK, so this is fine

src/client/async_client.rs
```diff
@@ -32,6 +43,133 @@ pub struct Client {
...
+        let payload = match serde_json::to_string(&inner_event) {
+            Ok(p) => p,
+            Err(e) => {
+                debug!(error = %e, "failed to serialize $feature_flag_called event");
+                return;
+            }
+        };
+        let http_client = self.http_client.clone();
+        let url = self.capture_url.clone();
+        tokio::spawn(async move {
```
dustinbyrne: Might be worth validating that a tokio runtime is still available in the current context.

Otherwise, `evaluate_flags` could be called from inside the tokio runtime and the resulting snapshot consumed from a context/thread without tokio.

Seems edge case-y, so I'll let you decide what the right thing to do here is.

src/lib.rs
```diff
@@ -26,6 +27,10 @@ pub use error::Error;
 pub use event::Event;
 
 // Feature Flags
+pub use feature_flag_evaluations::{
+    EvaluateFlagsOptions, EvaluatedFlagRecord, FeatureFlagEvaluations, FeatureFlagEvaluationsHost,
```
dustinbyrne: `EvaluatedFlagRecord`, `FlagCalledEventParams`, `FeatureFlagEvaluationsHost` look like they're private implementation details

src/client/mod.rs
```diff
@@ -76,6 +76,12 @@ pub struct ClientOptions {
     #[builder(default = "false")]
     local_evaluation_only: bool,
 
+    /// Whether to emit warnings for misuse of `FeatureFlagEvaluations` filter
+    /// helpers (e.g. calling `only_accessed()` before any access, or `only(...)`
+    /// with unknown keys). Set to `false` to silence these warnings.
+    #[builder(default = "true")]
+    feature_flags_log_warnings: bool,
```
dustinbyrne: same as https://github.com/PostHog/posthog-ruby/pull/137#discussion_r3170165738

dustinbyrne (commented):
Made comments on the async client, but some of these will apply to the blocking client as well

dustinbyrne (approved):

PR #109, opened 2026-05-05
fix: forward groups and group_properties to evaluate_all_flags

patricio-posthog:
## Motivation and Context

The release pipeline for v0.7.0 failed because `Client::evaluate_flags` (added in #105) calls `LocalEvaluator::evaluate_all_flags` with the old 2-argument signature, while #107 changed that method to require `groups` and `group_properties` as well.

The two PRs didn't conflict textually — different files, non-overlapping diffs — so neither PR's CI caught the type-level incompatibility. The repo doesn't run CI on `push: main`, so the failure surfaced only at release time when `cargo publish --dry-run` tried to compile the published tarball.

## Fix

`EvaluateFlagsOptions` already carries `groups` and `group_properties`. Forward them through to `evaluate_all_flags` from both the async and blocking clients.

## How did you test it?

- `cargo build` clean
- `cargo build --features e2e-test` clean (this is the path that fails in the release pipeline)
- `cargo test` — all 119 tests pass
- `cargo clippy --all-targets` clean

## Follow-up worth considering

To prevent this class of bug, the repo could add either:
- A required "branch up-to-date with `main`" rule on PRs (forces a rebase, which would have surfaced the type error before merge).
- A `push: main` build job mirroring the PR build (would catch incompatibilities post-merge before release).

### DIFF
--- a/.sampo/changesets/group-and-mixed-targeting-local-eval.md
+++ b/.sampo/changesets/group-and-mixed-targeting-local-eval.md
@@ -1,0 +1,7 @@
+---
+cargo/posthog-rs: minor
+---
+
+feat(flags): support group-targeted and mixed-targeting feature flags in local evaluation
+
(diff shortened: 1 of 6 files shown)


marandaneto:
rollback this commit https://github.com/PostHog/posthog-rs/commit/dfa3b3bbaab59c6438b96679154834074c542e65 so we can publish that version again

marandaneto (approved):

PR #116, opened 2026-05-28
Commit initial framework for SDK test harness integration

eli-r-ph:

Integrate the [SDK test harness](https://github.com/PostHog/posthog-sdk-test-harness) library into `posthog-rs`, including SDK wrapper for capture V0 and V1 test suite, Docker-based test harness, Actions etc.

The "legacy" capture suite should run as expected, with compliant functionality passing. The v1 suite will fail at first: I'm stubbing a 5xx-only endpoint and will implement the new v1 capture submission functionality in a follow on PR.

The new test harness integration will land initially without being required in CI; we can enable that in a follow on once it's smoke-tested as well

Locally and in CI

## :pencil: Checklist

- [ ] I reviewed the submitted code.
- [ ] I added tests to verify the changes.

### If releasing new changes

### DIFF
--- a/compliance/adapter/src/main.rs
+++ b/compliance/adapter/src/main.rs
@@ -1,0 +1,278 @@
+use std::sync::Arc;
+
+use axum::{
+    extract::State,
+    http::StatusCode,
+    response::IntoResponse,
+    routing::{get, post},
+    Json, Router,
+};
+use serde::{Deserialize, Serialize};
+use tokio::sync::Mutex;
+
+use posthog_rs::{CaptureMode, Client, ClientOptionsBuilder, Event};
+
+#[derive(Clone)]
+struct AppState {
+    inner: Arc<Mutex<AdapterState>>,
+    capture_mode: CaptureMode,
+}
+
+struct AdapterState {
+    client: Option<Client>,
+    total_events_captured: u64,
+    total_events_sent: u64,
+    total_retries: u64,
+    last_error: Option<String>,
+    requests_made: Vec<RequestRecord>,
+    pending_events: i64,
+}
+
+impl Default for AdapterState {
+    fn default() -> Self {
+        Self {
+            client: None,
+            total_events_captured: 0,
+            total_events_sent: 0,
+            total_retries: 0,
+            last_error: None,
+            requests_made: Vec::new(),
+            pending_events: 0,
+        }
+    }
+}
+
+#[derive(Serialize, Clone)]
+struct RequestRecord {
+    timestamp_ms: u64,
+    status_code: u16,
+    retry_attempt: u32,
+    event_count: usize,
+    uuid_list: Vec<String>,
+}
+
+#[derive(Deserialize)]
+struct InitRequest {
+    api_key: String,
+    host: String,
+    #[allow(dead_code)]
+    #[serde(default)]
+    flush_at: Option<u32>,
+    #[allow(dead_code)]
+    #[serde(default)]
+    flush_interval_ms: Option<u64>,
+    #[allow(dead_code)]
+    #[serde(default)]
+    max_retries: Option<u32>,
+    #[allow(dead_code)]
+    #[serde(default)]
+    enable_compression: Option<bool>,
+}
+
+#[derive(Deserialize)]
+struct CaptureRequest {
+    distinct_id: String,
+    event: String,
+    #[serde(default)]
+    properties: Option<serde_json::Value>,
+    #[serde(default)]
+    timestamp: Option<String>,
+}
+
+#[derive(Serialize)]
+struct HealthResponse {
+    sdk_name: &'static str,
+    sdk_version: &'static str,
+    adapter_version: &'static str,
+    capabilities: Vec<&'static str>,
+}
+
+#[derive(Serialize)]
+struct StateResponse {
+    pending_events: i64,
+    total_events_captured: u64,
+    total_events_sent: u64,
+    total_retries: u64,
+    last_error: Option<String>,
+    requests_made: Vec<RequestRecord>,
+}
+
+fn parse_capture_mode() -> CaptureMode {
+    match std::env::var("CAPTURE_MODE").as_deref() {
+        Ok("v1") => CaptureMode::V1,
+        _ => CaptureMode::V0,
+    }
+}
+
+async fn health(State(state): State<AppState>) -> Json<HealthResponse> {
+    let capability = match state.capture_mode {
+        CaptureMode::V0 => "capture_v0",
+        CaptureMode::V1 => "capture_v1",
+    };
+    Json(HealthResponse {
+        sdk_name: "posthog-rs",
+        sdk_version: env!("CARGO_PKG_VERSION"),
+        adapter_version: "0.1.0",
+        capabilities: vec![capability],
+    })
+}
+
+async fn init(
+    State(state): State<AppState>,
+    Json(req): Json<InitRequest>,
+) -> impl IntoResponse {
+    let mut s = state.inner.lock().await;
+
+    // Reset state
+    *s = AdapterState::default();
+
+    let options = ClientOptionsBuilder::default()
+        .api_key(req.api_key)
+        .host(req.host)
+        .capture_mode(state.capture_mode)
+        .build();
+
+    match options {
+        Ok(opts) => {
+            let client = posthog_rs::client(opts).await;
+            s.client = Some(client);
+            Json(serde_json::json!({ "success": true })).into_response()
+        }
+        Err(e) => {
+            let msg = e.to_string();
+            s.last_error = Some(msg.clone());
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(serde_json::json!({ "error": msg })),
+            )
+                .into_response()
+        }
+    }
+}
+
+async fn capture_event(
+    State(state): State<AppState>,
+    Json(req): Json<CaptureRequest>,
+) -> impl IntoResponse {
+    let mut s = state.inner.lock().await;
+
+    let client = match &s.client {
+        Some(c) => c,
+        None => {
+            return (
+                StatusCode::BAD_REQUEST,
+                Json(serde_json::json!({ "error": "SDK not initialized" })),
+            )
+                .into_response();
+        }
+    };
+
+    let mut event = Event::new(req.event, req.distinct_id);
+
+    if let Some(props) = req.properties {
+        if let Some(obj) = props.as_object() {
+            for (k, v) in obj {
+                let _ = event.insert_prop(k.clone(), v.clone());
+            }
+        }
+    }
+
+    if let Some(ts_str) = req.timestamp {
+        if let Ok(ts) = chrono::DateTime::parse_from_rfc3339(&ts_str) {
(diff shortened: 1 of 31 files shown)


jose-sequeira:
Should use the harness in parallel mode, so that harness tests run faster.

https://github.com/PostHog/posthog-sdk-test-harness/pull/10

src/client/async_client.rs
```diff
@@ -295,6 +303,20 @@ impl Client {
...
+        })
+    }
+
+    async fn capture_batch_v1(&self, _events: Vec<Event>) -> Result<(), Error> {
+        Err(Error::ServerError {
+            status: 500,
+            message: "Capture V1 not yet implemented".to_string(),
+        })
+    }
+
```
eli-r-ph, line 473: TODO: consolidate these; all v1 requests are batch

compliance/adapter/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+        }
+    }
+}
+
+async fn capture_event(
+    State(state): State<AppState>,
+    Query(params): Query<TestIdParam>,
+    Json(req): Json<CaptureRequest>,
+) -> impl IntoResponse {
+    let mut instances = state.instances.lock().await;
```
jose-sequeira: Seems like this mutex will actually make request serial.

You could store Arc<Client> in AdapterState, under the lock, you clone the Arc, drop the lock, call .capture() on the Arc, re-acquire briefly for counter updates, this way you can make serial calls

src/endpoints.rs
```diff
@@ -28,6 +30,7 @@ impl Endpoint {
         match self {
             Endpoint::Capture => "/i/v0/e/",
             Endpoint::Batch => "/batch/",
+            Endpoint::CaptureV1 => "/i/v1/analytics/events/",
```
jose-sequeira: In lib.rs this is re-exported, and lib.rs has:
```// Public interface - any change to this is breaking!```

I'm wondering if we need to have ` #[non_exhaustive]` in order to allow for this, or this means that a major version update is required

File: compliance/adapter/src/main.rs [hunk 4]
```diff
@@ -0,0 +1,342 @@
...
+
+async fn health(State(state): State<AppState>) -> Json<HealthResponse> {
+    let capability = match state.capture_mode {
+        CaptureMode::V0 => "capture_v0",
+        CaptureMode::V1 => "capture_v1",
+    };
+    Json(HealthResponse {
+        sdk_name: "posthog-rs",
+        sdk_version: env!("CARGO_PKG_VERSION"),
+        adapter_version: "0.3.0",
```
jose-sequeira: Here it is hardocded, but it is mismatched with Cargo.toml

compliance/adapter/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+    // EventOptions wiring (set_options) is applied in the V1 capture
+    // implementation branch where EventOptions exists on Event.
+    let _ = req.options;
+
+    let uuid = uuid::Uuid::now_v7().to_string();
+
+    match client.capture(event).await {
+        Ok(()) => {
+            s.total_events_captured += 1;
+            s.pending_events += 1;
```
jose-sequeira: Pretty sure pending_events manipulation just doesn't work here, since await completes before the counters are even touched?

jose-sequeira (commented):
Mostly looks good to me, only not approving since the Enum change is breaking and needs a clear fix

src/endpoints.rs
eli-r-ph: These are stubs, the 2nd in the stack adds `non_exchaustive` and a lot more nuance to the compliance suites and the posthog-rs impl v1 👍 

I'll check these against that

compliance/adapter/src/main.rs
eli-r-ph: hmm went over this concern with Claude during 1st iteration and thought we fixed this I'll have another look, thanks!

compliance/adapter/src/main.rs
eli-r-ph: fixed with locking changes (and this is redesigned somewhat as I added more robust batch tests in the next stacked PR, but same style of fix is now applied there too)

src/endpoints.rs
eli-r-ph: updated (and 2nd PR follows suit in an expanded as it adds full v1 impl)

compliance/adapter/src/main.rs
eli-r-ph: Ah this was fixed in the follow on PR where batch handling gets more robust after v1 API impl is landed. I also updated here to make this stage of the compliance suites more obviously correct, but full view of final state will be in 2nd PR

marandaneto:
Will review tomorrow 1st thing 👍

src/client/async_client.rs
```diff
@@ -309,18 +294,37 @@ impl Client {
...
+        }
+        request
+    }
+
+    async fn capture_v1(&self, _events: Vec<Event>) -> Result<(), Error> {
+        Err(Error::ServerError {
+            status: 500,
+            message: "Capture V1 not yet implemented".to_string(),
+        })
+    }
```
marandaneto: we should not release the SDK as is since one can now:

`options.capture_mode = CaptureMode::V1`

src/client/mod.rs
```diff
@@ -78,6 +89,16 @@ pub struct ClientOptions {
     #[builder(default = "false")]
     local_evaluation_only: bool,
 
+    /// Selects the capture pipeline. Defaults to `V0` (legacy `/batch/`).
+    #[builder(default)]
+    pub(crate) capture_mode: CaptureMode,
```
marandaneto: should we call this something more obvious like `endpoint` or something? `capture_mode` isnt clear tbh, i'd think its something related to capture behavior

or even, should we actually not expose this? Is the goal to keep the SDK with both endpoint options? Shouldn't we kill /batch for good and just fully migrate to v1 instead

src/client/mod.rs
```diff
@@ -78,6 +89,16 @@ pub struct ClientOptions {
...
+
+    /// Extra HTTP headers injected into every outbound capture request.
+    /// Used by the SDK test harness adapter to attach `X-Test-Id` for
+    /// parallel test isolation.
+    #[builder(default, setter(strip_option))]
+    pub(crate) extra_capture_headers: Option<HashMap<String, String>>,
```
marandaneto: how can we do this without leaking to a public API?

src/event_v1.rs
```diff
@@ -0,0 +1,56 @@
...
+#![allow(dead_code)]
+
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+/// A single event in the V1 wire format.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct V1Event {
```
marandaneto, line 10: Is it possible to reuse the `Event` struct?
If we implement features like https://posthog.com/docs/libraries/js/usage#redacting-information-in-events
We'd need to expose the event struct, and having 2 types based on the endpoint, with a strong typing language, that would be hard to make it work nicely

marandaneto (dismissed):
mainly because of https://github.com/PostHog/posthog-rs/pull/116#discussion_r3339258398 since this PR is ready-to-review, has a changeset and adds a pub method that throws

src/client/async_client.rs
eli-r-ph: 👋 hey I added a Rust feature `"capture-v1"` to ensure no one can use this accidentally on upgrade until we release. The backend is also going live this week so it will be ready to support calls prior to release.

The feature is setup in the next PR in the stack - this one didn't need it yet b/c the v1 implementation is just a stub (responds 5xx.)

Let me know if you'd like me to pull in that feature to this PR too 👍

src/client/mod.rs
eli-r-ph: 🤔 I was assuming so far there would be a transition rather than committing each SDK fully to v1 without an escape valve if some end user hits trouble? Also gives us a chance to add support to more SDKs before officially releasing the endpoint etc.

I'm happy to go with a hard cutover to v1 if that seems like a better play. Maybe take a peek at the next PRs in the series to see how the two fit together here and let me know which cutover you prefer.

Another option - use the Rust feature _as_ the endpoint gating during transition? Maybe it's just the `CaptureMode` that's in the way?

src/event_v1.rs
eli-r-ph: It's tricky because `Event` isn't a trait and it's public.

One option: V1 structures aren't reexported for public use atm. features that perform redactions and mutations would be applied on the legacy `Event` before it is converted to V1 for submission?

I'll ponder this one a bit 👍

src/client/mod.rs
eli-r-ph: I'm gonna put this behind a specific Rust feature so it's just not there unless the compliance suites are running against this code. It's only needed to inject the test ID headers 👍 updates on the way

src/client/mod.rs
eli-r-ph: Gonna ditch CaptureMode and just use the Rust feature `"capture-v1"` to gate which endpoint we go to. Thats a hard cutover, but it won't happen by accident to existing users during the transition, and we can choose when we eliminate the feature (and v0 capture) after v1 release. 

Let me know if you hate it, otherwise updates on the way 👍

src/endpoints.rs
eli-r-ph: This is also effected (goes away) if we decide to do a hard cutover for the transition (see other threads)

src/event_v1.rs
eli-r-ph: I think the simplest is to rethink this as a hard cutover gated by the `"capture-v1"` feature entirely. Then we can transition some of these public structures more smoothly. Going to update this PR stack to do that, see how it looks

src/client/mod.rs
eli-r-ph: OK done, let me know what you think

src/client/mod.rs
eli-r-ph: done

src/client/async_client.rs
eli-r-ph: done

src/event_v1.rs
eli-r-ph: updated to hard cutover concept. I left the V1 structures for now - they are not exported and could change under the hood after the rollout if we want. The public interface is still `Event` and as before we only apply V1 internals to reshape the payload for submission after any end-user mutations are completed.

If you see anything that gives you pause, let me know, we can iterate some more 👍 thanks!

dustinbyrne (approved):
This seems reasonable to approve given that the stubbed logic is behind a feature flag and the public API doesn't appear to change anywhere.

Regarding the feature flags, I'd consider splitting `capture_v0` / `capture_v1` implementations into separate internal modules, given they look like two pretty distinct paths. That way we can minimize the amount of logic scattered around behind feature gated paths in `async_client` / `blocking`, while maintaining the same public API.

I focused mainly on the actual SDK changes as it looks like the compliance stuff is still somewhat in flight / easy to change and any issues in the adapter are pretty negligible in terms of risk.

PR #118, opened 2026-05-30
Add V1 capture wire types, compression, and error mapping

eli-r-ph:
First slice of the V1 capture work, split out of the original #118 for review. Adds the gated `capture-v1` feature, V1 event/response types, request-body compression, and 401/402 error mapping. No behavior change under default features. Pipeline and tests follow in stacked PRs.


src/compression.rs
```diff
@@ -0,0 +1,99 @@
...
+                let mut encoder = brotli::CompressorWriter::new(&mut out, 4096, 5, 22);
+                if let Err(e) = encoder.write_all(data).and_then(|_| encoder.flush()) {
+                    Err(e)
+                } else {
+                    Ok(())
+                }
+            }
+            .map(|_| out)
+        }
+        CaptureCompression::Zstd => zstd::stream::encode_all(data, 0),
```
marandaneto, line 38: do we want to support all those cases? i think all other SDKs only support gzip or plain iirc

marandaneto (approved):

src/compression.rs
eli-r-ph: For v1 we'd like more options for the end user, and v1 backend supports all of these. My feeling is, for SDKs that can, support the ones we can, and for the rest, make the options a subset of these 4. In the Rust case we can support them all. For others, a min of gzip should be possible

PR #124, opened 2026-06-02
feat: emit $is_server property on captured events

turnipdabeets:
Adds `$is_server: true` to every event captured by this SDK so PostHog ingestion can identify server-side events.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -67,26 +67,9 @@ pub struct ClientOptions {
     #[builder(default = "false")]
     disable_geoip: bool,
 
-    /// Whether events originate from a server-side environment.
-    ///
-    /// Defaults to `true`. When `true`, the SDK adds `$is_server: true` to every
-    /// captured event so PostHog can identify server-side events. Set this to
-    /// `false` when running posthog-rs as a client/CLI (for example a desktop or
-    /// command-line tool) so the event's device OS is attributed normally and
-    /// `$is_server` is omitted. A caller-set `$is_server` property on an
-    /// individual event always takes precedence over this option.
-    ///
-    /// # Example
-    ///
-    /// ```ignore
-    /// use posthog_rs::ClientOptionsBuilder;
-    ///
-    /// let options = ClientOptionsBuilder::default()
-    ///     .api_key("your-api-key".to_string())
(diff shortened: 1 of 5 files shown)


src/client/mod.rs
```diff
@@ -67,6 +67,29 @@ pub struct ClientOptions {
...
+    /// use posthog_rs::ClientOptionsBuilder;
+    ///
+    /// let options = ClientOptionsBuilder::default()
+    ///     .api_key("your-api-key".to_string())
+    ///     .is_server(false) // running as a CLI/client
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    #[builder(default = "true")]
+    is_server: bool,
```
eli-r-ph, line 74: 🤔 optional: may want to trim down the comment and builder example on the new field it's pretty detailed

src/event.rs
```diff
@@ -134,7 +134,13 @@ pub struct InnerEvent {
...
 impl InnerEvent {
-    pub fn new(event: Event, api_key: String) -> Self {
+    /// Build an [`InnerEvent`] from an [`Event`], stamping the SDK metadata
+    /// (`$lib`, `$lib_version`, and so on) and, when `is_server` is `true`,
+    /// marking the event as originating from a server-side SDK via
+    /// `$is_server: true`. A caller-set `$is_server` property is always
+    /// preserved. The `is_server` flag is sourced from the client's
+    /// [`ClientOptions::is_server`](crate::ClientOptions) configuration.
+    pub fn new(event: Event, api_key: String, is_server: bool) -> Self {
         let uuid = event.uuid;
```
eli-r-ph, line 140: 🤔 it's a bit more duplicative, but I'd consider weaving in `is_server` the way `disable_geoip` is set now in this SDK instead of passing it through `InnerEvent::new` since it's just an Event prop.

eli-r-ph (approved):
Couple of small callouts to consider, but this LGTM 🚀 

The capture v1 stacked PRs rework the pattern for tucking things like `is_server` into event capture so its not too critical how we get this done at this stage

src/event.rs
turnipdabeets: updated!

PR #128, opened 2026-06-04
fix: bring v0 capture into SDK compliance

eli-r-ph:

The v0 SDK-harness compliance suite had 14 failing tests in three buckets: the wire field was `$distinct_id` (harness wants `distinct_id`), the v0 capture paths had no retry loop, and gzip was gated behind `capture-v1`. This fixes all three — **v0 is now 29/29, v1 stays 94/94.**

_Note: the feature-flags scoped compliance tests are still failing. These can be handled in another follow-on PR_

**`distinct_id` at root (no `$`) is legal and supported by the backend in v0.** `rust/capture` accepts `distinct_id` as the canonical field and only tolerates `$distinct_id` via a legacy serde alias, then normalizes to `distinct_id` on the Kafka envelope:

- `RawEvent.distinct_id` is the canonical field; `$distinct_id` is just a serde `alias`: [rust/common/types/src/event.rs#L67-L68](https://github.com/PostHog/posthog/blob/6b5d8897db037b97856f522bf72e8083b553838a/rust/common/types/src/event.rs#L67-L68)
- v0 `extract_distinct_id` tests assert both spellings extract identically: [rust/capture/src/v0_request.rs#L512-L513](https://github.com/PostHog/posthog/blob/6b5d8897db037b97856f522bf72e8083b553838a/rust/capture/src/v0_request.rs#L512-L513)

So this is "stop depending on a legacy alias and emit the canonical field," not a payload that was losing data — zero risk to ingestion.

**gzip uses a v0-compatible query param.** Capture reads compression from the `compression` query param (`gzip`/`gzip-js`) or sniffs gzip magic bytes — it does **not** consult `Content-Encoding`. So the v0 path sends both `?compression=gzip` (for the backend) and `Content-Encoding: gzip` (for the harness assertion / HTTP correctness):

- accepted query values: [rust/capture/src/payload/types.rs#L28](https://github.com/PostHog/posthog/blob/6b5d8897db037b97856f522bf72e8083b553838a/rust/capture/src/payload/types.rs#L28)
- decode path (query hint OR magic bytes): [rust/capture/src/payload/decompression.rs#L116](https://github.com/PostHog/posthog/blob/6b5d8897db037b97856f522bf72e8083b553838a/rust/capture/src/payload/decompression.rs#L116)

429 is retried only when it carries `Retry-After` (a bare 429 stays a terminal rate-limit, preserving existing behavior); v1 never receives 429, so that widening is scoped to v0.

- New unit tests: wire-shape (`distinct_id` at root, single + batch), retry predicates, gzip roundtrip.
- New integration tests (`httpmock`): retryable vs terminal statuses, attempt counts, `Retry-After`, byte-identical resends, and gzip (header + query param + body decodes to the event).
- `cargo test` green on default, `capture-v1`, blocking, and blocking+`capture-v1`; `cargo clippy -- -D warnings` clean.
- Compliance suites: **v0 29/29**, **v1 94/94**.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ tokio = { version = "1", features = [
     "time",
     "macros",
 ], optional = true }
-flate2 = { version = "1.0", optional = true }
+flate2 = "1.0"
 brotli = { version = "7.0", optional = true }
 zstd = { version = "0.13", optional = true }
 
@@ -38,6 +38,7 @@ zstd = { version = "0.13", optional = true }
 dotenv = "0.15.0"
 ctor = "0.1.26"
 tokio = { version = "1", features = ["full"] }
+flate2 = "1.0"
 httpmock = "0.7"
 serde_json = "1.0"
 futures = "0.3"
@@ -47,7 +48,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 default = ["async-client"]
 e2e-test = []
 async-client = ["tokio"]
-capture-v1 = ["flate2", "brotli", "zstd"]
+capture-v1 = ["brotli", "zstd"]
 test-harness = []
 
 [workspace]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,6 @@ zstd = { version = "0.13", optional = true }
 dotenv = "0.15.0"
 ctor = "0.1.26"
 tokio = { version = "1", features = ["full"] }
-flate2 = "1.0"
 httpmock = "0.7"
 serde_json = "1.0"
 futures = "0.3"
--- a/src/client/async_client.rs
+++ b/src/client/async_client.rs
@@ -343,21 +343,9 @@ impl Client {
         super::v0_capture::prepare_event(&mut event, &defaults);
         let payload =
             super::v0_capture::build_capture_payload(event, self.options.api_key.clone())?;
-
         let url = self.options.endpoints().build_url(Endpoint::Capture);
-        let request = self
-            .client
-            .post(&url)
-            .header(CONTENT_TYPE, "application/json")
-            .body(payload);
-        let request = super::v0_capture::apply_extra_headers(&self.options, request);
-
-        let response = request
-            .send()
-            .await
-            .map_err(|e| Error::Connection(e.to_string()))?;
-
-        check_response(response).await
+        let (body, gzip) = self.encode_v0_body(payload);
+        self.send_v0_with_retry(&url, body, gzip).await
     }
 
     #[cfg(not(feature = "capture-v1"))]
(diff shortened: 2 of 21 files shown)


eli-r-ph:
Addressed both Greptile P2s in d7d4e60:
- extracted the duplicated `encode_v0_body` into `v0_capture::encode_body` (single source for the gzip fallback path)
- consolidated the three `parse_retry_after` tests into one table-driven case

src/client/async_client.rs
```diff
@@ -374,19 +362,63 @@ impl Client {
...
-            .await
-            .map_err(|e| Error::Connection(e.to_string()))?;
+    /// POST `body` to `url`, retrying transient failures (transport errors and
+    /// 408/429/500/502/503/504) up to `max_capture_attempts`. The body is built
+    /// once by the caller and resent byte-for-byte, so a retried event keeps
+    /// its UUID and timestamp — which dedup relies on. When `gzip`, the request
+    /// advertises `Content-Encoding: gzip` and the `compression=gzip` query
+    /// param (capture reads the query param on v0, not the header).
+    #[cfg(not(feature = "capture-v1"))]
+    async fn send_v0_with_retry(&self, url: &str, body: Vec<u8>, gzip: bool) -> Result<(), Error> {
```
nickbest-ph: duplicated in `src/client/blocking.rs`

might wanna consider DRYing that up

Cargo.toml
```diff
@@ -30,14 +30,15 @@ tokio = { version = "1", features = [
...
-flate2 = { version = "1.0", optional = true }
+flate2 = "1.0"
 brotli = { version = "7.0", optional = true }
 zstd = { version = "0.13", optional = true }
 
 [dev-dependencies]
 dotenv = "0.15.0"
 ctor = "0.1.26"
 tokio = { version = "1", features = ["full"] }
+flate2 = "1.0"
```
nickbest-ph: nit: if this is now a non-optional dependency, does it need to be in dev-dependencies?

src/client/async_client.rs
```diff
@@ -374,19 +362,63 @@ impl Client {
...
+                Ok(response) => {
+                    let status = response.status().as_u16();
+                    let retry_after = super::retry::parse_retry_after(response.headers());
+                    if !(super::retry::should_retry_v0(status, retry_after.is_some())
+                        && attempt < self.options.max_capture_attempts)
+                    {
+                        return check_response(response).await;
+                    }
+                    let delay =
+                        super::retry::backoff_duration(&self.options, attempt + 1, retry_after);
```
nickbest-ph: attempt starts at 1. is it on purpose that the first back off delay gets attempt = 2 passed into it?

nickbest-ph (approved):

src/client/async_client.rs
eli-r-ph: ah yes thanks the pkg split left a few of these, thought I cleaned them up I'll take another pass thanks! 🙇

src/client/async_client.rs
eli-r-ph: ah thats a divergence in the old v0 code that should have it's own test case, v1 handles this. updated! 🙇

Cargo.toml
eli-r-ph: yeah that is weird, no need for it. removed 🙇 and consolidated the compression handling to simplify this

PR #129, opened 2026-06-08
feat: add manual rust error tracking

cat-ph:

Adds a manual Error tracking v1 for the Rust SDK behind the `error-tracking` feature. The API supports building exception payloads from Rust errors, capturing exception events, optional stacktrace capture via the `backtrace` crate, and client-level stacktrace/in-app frame options.

- `cargo generate-lockfile`
- `cargo fmt --all -- --check`
- `cargo check --features error-tracking`
- `cargo clippy --all-targets --features error-tracking -- -D warnings`
- `cargo test --features error-tracking`
- `cargo test --no-default-features --features error-tracking`

also went through with a local project sending a test stacktrace:

CleanShot 2026-06-08 at 22 51 43

## :pencil: Checklist

### If releasing new changes


src/client/blocking.rs
```diff
@@ -287,6 +291,45 @@ impl Client {
         self.capture_v0(event)
     }
 
+    /// Capture a collection of events with a single request.
+    /// Capture an exception event, sending it to PostHog Error Tracking.
+    #[cfg(feature = "error-tracking")]
+    pub fn capture_exception(&self, exception: ExceptionCapture) -> Result<(), Error> {
```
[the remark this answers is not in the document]
cat-ph: Fixed in 1cb4436. Removed the stale batch-capture doc line from capture_exception.

src/error_tracking.rs
```diff
@@ -0,0 +1,829 @@
...
+    }
+    trim_to_max_frames(&mut frames, options.max_frames());
+
+    ExceptionStacktrace::raw(frames)
+}
+
+fn is_internal_capture_frame(function: &str) -> bool {
+    function.starts_with("backtrace::")
+        || function.contains("DefaultStackTraceExtractor::capture")
+        || function.contains("capture_frames_current_first")
```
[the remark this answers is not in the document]
cat-ph: Fixed in 1cb4436. Removed the stale DefaultStackTraceExtractor predicate from the internal-frame filter.

src/client/async_client.rs
```diff
@@ -303,6 +307,45 @@ impl Client {
         self.capture_v0(event).await
     }
 
+    /// Capture a collection of events with a single request.
+    /// Capture an exception event, sending it to PostHog Error Tracking.
+    #[cfg(feature = "error-tracking")]
+    pub async fn capture_exception(&self, exception: ExceptionCapture) -> Result<(), Error> {
```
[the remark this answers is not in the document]
cat-ph: Fixed in 1cb4436. Removed the stale batch-capture doc line from capture_exception.

Cargo.toml
```diff
@@ -49,6 +51,7 @@ e2e-test = []
 async-client = ["tokio"]
 capture-v1 = ["brotli", "zstd"]
 test-harness = []
+error-tracking = ["dep:backtrace"]
```
cat-ph, line 54: this is mostly to prevent a dependency on backtrace above if not using ET

examples/error_tracking.rs
```diff
@@ -0,0 +1,31 @@
...
+    let host = std::env::var("POSTHOG_HOST").unwrap_or_else(|_| posthog_rs::DEFAULT_HOST.into());
+    let client = client((api_key.as_str(), host.as_str())).await;
+
+    let error = std::io::Error::other("checkout failed");
+    let exception = client
+        .exception_from_error(&error)
+        .with_distinct_id("user-123")
+        .with_prop("route", "/checkout")?;
+
+    client.capture_exception(exception).await?;
```
marandaneto: that is different than the other SDKs

other SDKs would do:

`client.capture_exception(error).await?;`
and internally `capture_exception` calls `exception_from_error` so `exception_from_error` is private and not a public API
`with_distinct_id` should either be added automatically or be an optional param for `capture_exception`
`with_prop` should be an optional param for `capture_exception`

wdyt?

src/client/async_client.rs
```diff
@@ -233,6 +237,60 @@ impl Client {
...
+    }
+
+    /// Build an exception capture from a Rust error using this client's Error Tracking options.
+    #[cfg(feature = "error-tracking")]
+    pub fn exception_from_error<E>(&self, error: &E) -> ExceptionCapture
+    where
+        E: StdError + ?Sized,
+    {
+        ExceptionCapture::from_error_with_options(error, self.options.error_tracking())
+    }
```
marandaneto: i dont think this should be public

src/client/async_client.rs
```diff
@@ -233,6 +237,60 @@ impl Client {
...
+
+        let exception =
+            ExceptionCapture::from_error_with_options(error, self.options.error_tracking())
+                .with_distinct_id(distinct_id);
+        self.capture_exception(exception).await
+    }
+
+    /// Capture a Rust error personlessly, sending it to PostHog Error Tracking.
+    #[cfg(feature = "error-tracking")]
+    pub async fn capture_error_anon<E>(&self, error: &E) -> Result<(), Error>
```
marandaneto: we should make the distinct_id as optional and then we dont need this method

README.md
```diff
@@ -19,7 +20,7 @@ Add `posthog-rs` to your `Cargo.toml`.
 
 ```toml
 [dependencies]
-posthog-rs = "0.3.7"
+posthog-rs = "0.8.0"
```
marandaneto:
```suggestion
posthog-rs = "$version"
```

otherwise this is always outdated

marandaneto:
how does auto capture work here? can we hook up the PanicHookInfo?

examples/error_tracking.rs
cat-ph: yeahhh thanks @marandaneto  - I wasn't very sure of the best approach here. the issue is we can't really do optional args, well actually we could send Option<..>, I tried this but was pretty weird cause it still forces you to send None, which felt a bit worse. I saw we use the same pattern for Event::new vs Event::new_anon, that's why I separated those like that

in this case capture_error was a 'helper' for a rust `error` explicitly, and Exception is our exception "type" builder style where we can add properties/flags etc

cat-ph:
yep for auto capture, I have a separate draft stacked PR, was just testing it now: https://github.com/PostHog/posthog-rs/pull/130

src/client/async_client.rs
cat-ph: I kinda followed `Event::new` & `Event::new_anon` here since it seems more rust-y, were you thinking of passing an `Option<String>` and forcing callers to send None to make it explicit they're not sending a distinct_id?

examples/error_tracking.rs
cat-ph: not very sure how we can do the overload here for making `exception_from_error` private, I'll try locally with traits/generics to see how muddy it gets

we _could_ lean on ExceptionCapture::from_error and capture_exception separately, I think this goes back to the API 'decision' above, my current approach/intent was:

`capture_error(&error, distinct_id)` happy path for a rust Error
`capture_error_anon(&error)` personless error
`capture_exception(ExceptionCapture)` for a builder-style path to adding custom props/flags/groups etc

marandaneto: yeah no method overloads, and optional forces passing None

another approach is:

`capture_exception(exception, captureOptions)`

where `captureOptions` is an object with a builder pattern that you can pass the optional `distinct_id`, the optinal props, etc, but we should hide the `exception_from_error` bits since its internal and the SDK can do this automatically

src/client/async_client.rs
marandaneto: see https://github.com/PostHog/posthog-rs/pull/129#discussion_r3382383484

yeah i get because of new and new_anon

if i'd have designed, i'd have done like this

```
Event::new(event_name)
    .distinct_id(x)
    .anonymous(true)
// etc
```

or

```
captureOptions = captureOptions()
.anonymous(true)

capture(event, captureOptions)
```

feels more idiomatic and the public API is smaller and we can expand the captureOptions easily instead of having a new public method for every alternative

marandaneto:
What if the final bundle is optimized and stripped of debug symbols? Do we intend to offer a tool for uploading debug symbols and symbolication? not a blocker since this can be done later

cat-ph:
absolutely, that's up next, I have to add support in cymbal too, and then want to add uploading debug symbols + symbolication 😁

examples/error_tracking.rs
cat-ph: yeah I agree about hiding exception_from_error, I'll do that either way first!

I like the idea behind having a single captureOptions to  but I kinda dislike the rust-y way of doing it 😢 (because of needing to pass `CaptureOptions::default()` everytime or forcing `None` or `Some(captureOptions)`)

moving to the thread below

src/client/async_client.rs
cat-ph: hmmmmm. this also made me think, we could get away with an event constructor and just relying on it instead?

i.e.

```rs
Event::from_error(&error)
  .set_distinct_id(..) // or a from_error / from_error_anon until we change the new/new_anon dynamic
```

although then we don't have an explicit `capture_exception` while all other SDKs do...

-- 

alternatively, the (yet another) option of keeping the exception type (with exception specific things, like fingerprint etc) and having an `into_event_anon` and `into_event("distinct_id")`, we'd have something like:

```rs
impl Exception {
    pub fn into_event(self, distinct_id: impl Into<String>) -> Event { ... }
    pub fn into_event_anon(self) -> Event { ... }
}

client.capture_exception(Exception::from_error(&err, "distinct_id")).await?; // happy path
client.capture_exception(Exception::from_error_anon(&err)).await?; // happy, personless path

// or the other, longer path is going to event

let mut event = Exception::from_error(&err).into_event("user-123"); // or into_event_anon()
event.insert_prop("route", "/checkout")?;
event.with_flags(&flags);
event.add_group("company", "acme");
client.capture(event).await?; 
```

wdyt @marandaneto ?

cat-ph: actually my bad, I meant `from_error` and `from_error_anon` to keep happy path simpler, and just a simple `into_event()` in the last suggestion

cat-ph: tried out the `capture_exception` / `capture_exception_anon` sugar & keeping consistency with SDKs + `from_error`..`into_event` / `into_event_anon` to convert into event and the regular event stuff (in the latest commit, the others just fix the other stuff, version/public thingy we didn't want)

examples/error_tracking.rs
marandaneto: yeah its hard without method overloads and optionals with default params, same feeling

src/client/async_client.rs
marandaneto: i find the `into_event` (nor `from_error`) not a great DX

IMO public APIs should be self-explanatory, eg

`capture(event, everythingOptional)`
the same goes for an error
`capture_exception(error, everythingOptional)`

the `event` (string) and the `error` (rust error type) should be the raw object instead of relying on people figuring yet another API to construct the value (eg `Exception::from_error(&err)`)
i am not a rust expert so its hard for me to say
but i'd rather have a pubic API that accepts the raw `error` and then its ok to get more optional metadata that will be part of the final exception event

if we want to at least match all the other SDKs, i'd not overcomplicate too much and go simply with:

`capture(string event, captureOptions)`

or

`event = event(string event name).withSomething(x).withSomething(y)`
// or just `event = event(string event name)`
`capture(event obj)` where you can construct the `event` with many optional values (like a builder pattern)

i think we can apply the same approach for the error/exception

`exception = exception(error).withSomething(x).withSomething(y)`
// or just `exception = exception(error)`
`capture_exception(exception)` (and we do the exception_from_error internally)

or

`capture_exception|error(error, captureOptions)`

marandaneto: btw there are rust experts within posthog so feel free to chime them in, in case i am talking bs :D

hpouillot: I think `from_error` and `into_event` makes sense but I would use the standard trait for that

```rust
// Build pattern for exception
let exception = ExeceptionEvent::from(error).distinct_id("").prop("key", "value)

impl From<ExceptionEvent> for Event
impl From<Error> for ExceptionEvent

// Not sure this is the current API, but it would make things easier since we could convert different event shapes to the canonical form.
capture(impl Into<Event>)
and 
// a specific method for just capturing exceptions.
capture_exception(impl Into<ExceptionEvent>)

// So we could do:
capture_exception(err)
```

Wdyt ?

marandaneto: i find it strange that `capture_exception` captures an event type (Assuming Exception::from_error(&err).into_event("user-123"))
then we dont need `capture_exception` since one can just call `capture` instead
IMO the `capture_exception` should receive the raw error type, thats why its called `capture_exception`

now if `ExceptionEvent` and `Event` are different things, thats ok, but its much more friction for the user to capture an error
they have to figure out the `ExceptionEvent`, what to call for passing `error`, etc

IMO the most friendly API still is:

```rust
   client.capture_exception(&err).await?; // or capture_error

   client
       .capture_exception_with(
           &err,
           CaptureOptions::new() // builder
               .distinct_id("user-123")
               .property("plan", "pro")?
               .fingerprint("db-timeout")
               .level("warning"),
       )
       .await?;
 ```

 Recommended public API shape:

 ```rust
   pub async fn capture_exception<E>(&self, error: &E) -> Result<(), Error>
   where
       E: StdError + ?Sized;

   pub async fn capture_exception_with<E>(
       &self,
       error: &E,
       options: CaptureExceptionOptions,
   ) -> Result<(), Error>
   where
       E: StdError + ?Sized;
 ```

 And:

 ```rust
   #[derive(Clone, Debug, Default)]
   pub struct CaptureExceptionOptions {
       pub distinct_id: Option<String>,
       pub properties: HashMap<String, serde_json::Value>,
       pub groups: HashMap<String, String>,
       pub fingerprint: Option<String>,
       pub level: Option<String>,
       // future: tags, mechanism, feature flags, timestamp, etc.
   }
 ```

will leave that to you both that def know more rust than me :D

Comment (cat-ph) [reply src/client/async_client.rs]: heh I don't know a _ton_ of rust either 😆 thanks both! 💙 

I think 'decisions' after this discussion:

- do we want an explicit capture_exception? I'll lean into yes, to keep consistency with other SDKs
- which, if yes.. do we want std error, or a custom exception type? this goes into type semantics, if we do the trait thing and exception is a type of event, then it'll also be able to capture_exception(events) (what Manoel pointed), so I'm inclined to say std error only
- with that in mind, we need to allow capturing extra props (ideally) without duplicating stuff in events, and here we have the two choices:
  - @marandaneto's suggested capture_exception_with and a separate builder, that we later on can adapt & use in event too
  - @hpouillot's exception builder directly, which feels more rust-y but then the capture_exception is the discussion above

based on that, I quite like the `capture_exception_with` proposal, with options that we can separately use with events too etc, it feels pretty nice! we can also extend it later to what Hugues suggested to make an `impl Into<Exception>`

^ only small tradeoff is, the happy path is kinda personless, but I'd say we can make it work. we could even _keep_ the `capture_exception_anon(&err)` and `capture_exception(&err, distinct_id)` and add `capture_exception_with` but might be a bit excessive/confusing 😆 

I'll draft the capture_exception_with change, unless y'all have any final thoughts on this personless asymmetry

Comment (marandaneto) [reply src/client/async_client.rs]: this is a stateless SDK, so its ok that the happy path is w/o a distinct id (imo)
thats why we'd have the `capture_exception_with` which one can set distinct id and more

Comment (hpouillot) [reply src/client/async_client.rs]: Yeah I agree, `capture_exception` could take the error directly since it implements Into<ExceptionEvent>. So you could give it either an error directly or an ExceptionEvent since both impl the trait. We could also potentially take a string as the error message and construct an ExceptionEvent for that which aligns with the rest of the SDKs

Comment (cat-ph) [reply src/client/async_client.rs]: yep, I think for now users can just create an error from string to keep it simpler, and we can (non-breaking) convert the `From/Into<Exception>` in a separate PR (I think there's some extra discussion since it'd break some impls like `From<&TheirError> for Exception`)

updated!

Comment (marandaneto) [reply src/client/async_client.rs]: cool, i know some other people do like capture_message(str) instead so its clear (error vs a string)
the dif is that depending on the ecosystem you can also throw a string, you cannot in rust, i think you can panic with a string tho

Comment (hpouillot) [reply src/client/async_client.rs]: You cannot throw at all in rust, but the error could be a string Result<Value, String>

Comment (cat-ph) [reply src/client/async_client.rs]: we could also similarly do `capture_exception_message(str)` to make it clear that: it's captured as an exception (similar to string captureException in other SDKs), and it takes a custom string

Comment (cat-ph) [reply src/client/async_client.rs]: I _think_ the current shape is good to go now, piggybacked on the v1 slack thread to add Dylan & Olly in case they have some thoughts too 😆

L.E. and Eli too since I saw recent merges here 😁

File: src/error_tracking.rs [hunk 9]
```diff
@@ -0,0 +1,935 @@
...
+
+/// Error Tracking stacktrace and frame classification options.
+#[derive(Builder, Clone, Debug)]
+#[builder(default)]
+pub struct ErrorTrackingOptions {
+    capture_stacktrace: bool,
+    max_frames: usize,
+    max_error_sources: usize,
+    in_app_include_paths: Vec<String>,
+    in_app_exclude_paths: Vec<String>,
```
marandaneto, line 57: lets add a /// comment explaining all public methods, configs and event properties that are public
if possible with examples, LLMs love that, so agents dont trip and guess work

src/error_tracking.rs
```diff
@@ -0,0 +1,935 @@
...
+use crate::{Error, Event};
+
+const DEFAULT_MAX_FRAMES: usize = 64;
+const DEFAULT_MAX_ERROR_SOURCES: usize = 50;
+
+/// Error Tracking stacktrace and frame classification options.
+#[derive(Builder, Clone, Debug)]
+#[builder(default)]
+pub struct ErrorTrackingOptions {
+    capture_stacktrace: bool,
```
marandaneto, line 46: curious why this is optional? perfcost?

src/error_tracking.rs
```diff
@@ -0,0 +1,935 @@
...
+    max_frames: usize,
+    max_error_sources: usize,
```
marandaneto: i dont recall seeing this in other SDKs, are rust traces bigger than usual?

src/error_tracking.rs
marandaneto, line 57: would also be possible to set a crate name here as eeg exclude paths? eg all frames that come from 3rd party crate X should be excluded (inapp=false)

src/client/async_client.rs
marandaneto: i like it!

src/error_tracking.rs
cat-ph: yessir, I think `backtrace::trace` + per-symbol resolution can cost a bit.. this is kinda needed now until native symbolication works. (though useful afterwards too, since default `--release` builds keep the symbol tables so this also resolves the names, just not files/lines)

but good point actually, just realized the check for this happens _after_ the call site where we already build the frames 😅 fixing

src/error_tracking.rs
cat-ph: I think it's in other SDKs too, but not configurable (the android ones from a few weeks back that caused perf issues with resolution in cymbal), just an extra safety for longer chains since it's kinda common to just wrap with anyhow chains etc.

but it makes more sense to not have those as extra configurations, I agree, dropping those

src/error_tracking.rs
cat-ph: yep this already works, I'll add some examples, `in_app_exclude_paths(vec!["other_crate::"])` would do it

src/error_tracking.rs
marandaneto: yeah i understand, my point is that errors without a stacktrace are useless (isnt it?) so i dont see the value of having a config to disable it tbh
like i dont see how anyone would disable this and still find value in our product

src/error_tracking.rs
marandaneto: we dont have the `_paths` suffix in other SDKs but thats alright if it makes sense for the rust packages

src/error_tracking.rs
```diff
@@ -0,0 +1,935 @@
...
+        }
+    }
+
+    /// Set a custom exception fingerprint.
+    pub fn set_fingerprint<S: Into<String>>(&mut self, fingerprint: S) {
+        self.fingerprint = Some(fingerprint.into());
+    }
+
+    /// Set the exception severity level. Defaults to `"error"`.
+    pub fn set_level<S: Into<String>>(&mut self, level: S) {
```
hpouillot: we can set the level on capture options and exception ?

src/error_tracking.rs
```diff
@@ -0,0 +1,935 @@
...
+
+    backtrace::trace(|frame| {
+        if skipped < skip {
+            skipped += 1;
+            return true;
+        }
+
+        let mut pushed = false;
+        backtrace::resolve_frame(frame, |symbol| {
+            if pushed {
```
hpouillot: Don't we miss frames if we stop on the first one ?

src/event.rs
```diff
@@ -40,6 +40,13 @@ pub struct Event {
     uuid: Uuid,
     #[serde(skip)]
     options: EventOptions,
+    /// Pending Error Tracking payload. Client policy is applied and the
+    /// `$exception_*` properties are written at capture time, so an event
+    /// converted from an [`crate::error_tracking::Exception`] always honors
+    /// the capturing client's configuration.
+    #[cfg(feature = "error-tracking")]
+    #[serde(skip)]
+    pub(crate) exception: Option<crate::error_tracking::Exception>,
```
hpouillot: This looks off, do we really need to attach the exception on the event ? Can't we convert it into an event instead ?

src/error_tracking.rs
cat-ph: right, but technically this is stacktrace at the calling point, not the error creation one, so it's not necessarily that useful compared to other languages, this is just for adding some extra context. the more useful context is the source chain / error types/names etc (even without the stacktrace, which would only be on the very first item in the chain)

in some cases, people use anyhow to capture the origin stack when the error is constructed, which would be the "right" stack (though we need some sort of integration with anyhow to read that)

src/error_tracking.rs
cat-ph: a bit confusing but it's not duplicated, `Exception` isn't technically public, we use it in `build_exception_event` to set it based on the capture options 😁

src/event.rs
cat-ph: yeah indeed it's a bit odd, it was because of the `finalize_exception` flow from when we had `exception_to_event` before, but I think we can get rid of it now actually, good point, updating

src/error_tracking.rs
cat-ph: this is per actual frame (we want one symbol per physical frame, the callback happens for each inlined function at that address AIUI), the outer loop still continues

```
physical frame #1 → resolve_frame fires 7 times, in this order:
  [0] backtrace::libunwind::trace      ← innermost: the actual execution point   ← WE KEEP THIS
  [3] innermost_helper   (line 6)      ┐
  [4] middle_layer       (line 29)     │ the inline ancestry — each with its
  [5] outer_caller       (line 34)     │ exact call-site line number
  [6] main               (line 38)     ┘ outermost: the physical function        ← ALL DROPPED
```

(better example than I could explain 😬)

hpouillot: Yes, It is the same for other compiled languages, one frame could resolve to mulitple frames if the compiler inline them, so we probably want to keep them.

cat-ph: yep fair, took a look at apple & how we process in cymbal etc, that makes sense! updated 💙

src/error_tracking.rs
cat-ph: small change, I made them `pub (crate)` to make it a bit clearer they're not actually public public

README.md
```diff
@@ -46,6 +47,64 @@ if is_enabled {
...
+let client = client("your-api-key").await;
+let error = std::io::Error::new(std::io::ErrorKind::Other, "checkout failed");
+
+// Associate the exception with a person and attach optional context.
+client.capture_exception_with(
+    &error,
+    CaptureExceptionOptions::new()
+        .distinct_id("user-123")
+        .property("route", "/checkout").unwrap()
+        .fingerprint("checkout-error"),
```
hpouillot, line 68: Should we keep fingerprint for more advanced examples or explain the what it does ? I am wondering if agents will override everything otherwise

```suggestion
        .property("route", "/checkout").unwrap(),
```

hpouillot (approved):
🦀 🦀 🦀

README.md
hpouillot: Hum my bad, didn't see

https://github.com/PostHog/posthog-rs/pull/129/changes/BASE..330b04fb45e0fd3ba17684b4fdccae04e8a7d3f3#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R106

Cargo.toml
dustinbyrne: should it be default?

dustinbyrne (commented):
this is test failing when combined with the `capture-v1` feature - we should either add this combo to the matrix or I strongly suggest including this as a default feature

Cargo.toml
cat-ph: the only problem was with the `=0.3.74` pin on backtrace tbh, cause it's pretty restrictive.. we have `rust-version = "1.78.0"`, but looking at it, other deps already break this? and I don't think we have CI testing this msrv, so maybe not needed to pin this..

cat-ph: probably don't need to care about it rn, I think we need to bump to 1.85.0 anyway, we have `idna_adapter 1.2.2, getrandom 0.4.2, hashbrown 0.17.1, indexmap 2.14.0, security-framework 3.7.0` which fail below it

dropped the pin & made it default!

cat-ph:
yea good spot, updated, thanks!

dustinbyrne (approved):
🚀

marandaneto (approved):

README.md
```diff
@@ -46,6 +47,65 @@ if is_enabled {
 }
 ```
 
+## Error Tracking
```
marandaneto, line 50: @cat-ph too late to the party but see this comment

see https://github.com/PostHog/posthog-rs/pull/140
lets avoid docs in those files
lets move those docs to the oficial docs under https://posthog.com/docs/libraries/rust or the official error tracking rust docs
that should be the canonical docs

PR #130, opened 2026-06-08
feat: add panic autocapture

cat-ph:
## Summary
- add process-wide `install_panic_hook` panic autocapture for the `error-tracking` feature (now rebased onto main after #129 merged)
- panics are captured personlessly as `$exception` events: the panic payload as the exception value, panic-site `$exception_panic_file`/`_line`/`_column` properties from `PanicInfo::location()` (compiler-provided, survives release builds without debug info), and a call-site stacktrace honoring `capture_stacktrace`
- capture is best-effort and panic-safe: the event is built on the hook thread, then sent synchronously from a dedicated sender thread (panic count is thread-local, so a panic anywhere in the HTTP stack becomes a joinable thread panic instead of a process abort), bounded by `panic_capture_timeout_seconds` (default 2s); the previously installed hook always runs afterwards
- `before_send` hooks are deliberately bypassed for panic captures: hooks run arbitrary user code that cannot safely execute inside a panic hook (a hook mutex may be held mid-panic by this thread, and a nested panic aborts the process); capture failures go to stderr rather than tracing for the same reason
- installing the hook twice returns `Error::PanicHookAlreadyInstalled`

## Test plan
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` (default and `--features capture-v1`)
- `cargo test` across the four feature configs: default, `capture-v1`, `--no-default-features --features error-tracking`, `--no-default-features --features error-tracking,capture-v1`
- regression coverage: personless capture + previous-hook chaining, disabled-client no-send, panics on tokio runtime threads (multi-thread and `current_thread`), inline-frame expansion and internal-frame stripping


src/error_tracking.rs
```diff
@@ -434,6 +527,85 @@ fn is_internal_capture_frame(function: &str) -> bool {
...
+
+fn prepare_panic_event(event: &mut Event, defaults: &CaptureDefaults) {
+    if defaults.disable_geoip {
+        event.insert_prop_default("$geoip_disable", Value::Bool(true));
+    }
+    if defaults.is_server {
+        event.insert_prop_default("$is_server", Value::Bool(true));
+    }
+    event.prepare_for_v0();
+}
```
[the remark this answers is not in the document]
cat-ph: Thanks, this duplication is intentional for now. Panic autocapture is currently a temporary synchronous V0 sender because the hook runs during unwinding and cannot rely on the normal Capture V1 client path. Normal capture is migrating behind `capture-v1`, so I kept panic V0 prep local and added an inline comment instead of lifting V0 prep into feature-agnostic client code.

README.md
```diff
@@ -106,6 +108,17 @@ let options = ClientOptionsBuilder::default()
 | `$exception_level` | `"error"` unless set via `CaptureExceptionOptions::level`. |
 | `$exception_fingerprint` | Only present when set via `CaptureExceptionOptions::fingerprint`; overrides server-side issue grouping. |
 
+### Panic autocapture
```
marandaneto: just a side note

see https://github.com/PostHog/posthog-rs/pull/140
lets avoid docs in those files
lets move those docs to the oficial docs under https://posthog.com/docs/libraries/rust
that should be the canonical docs

cat-ph: woops, my bad. yeah, I was going to open a separate PR with docs for the manual capture + panic hook, but it makes sense to not pollute the readme 😅 I'll drop those

marandaneto:
i think the hook should be installed during SDK init
eg
     - init_global(...).with_panic_hook(...), or
     - ClientOptions::capture_panics(true)

instead of a standalone public API - users would need to call install_panic_hook manually,  they would need to find out etc

src/client/mod.rs
```diff
@@ -238,6 +243,18 @@ impl ClientOptions {
...
+    #[cfg(feature = "error-tracking")]
+    pub(crate) fn panic_capture_timeout_seconds(&self) -> u64 {
+        self.panic_capture_timeout_seconds
+    }
+
+    /// Project API key used by capture requests.
+    #[cfg_attr(not(feature = "error-tracking"), allow(dead_code))]
+    pub(crate) fn api_key(&self) -> &str {
+        &self.api_key
+    }
```
marandaneto, line 256: tbh i dont like this approach
i think this sdk should have a runtime-independent transport layer, where the panic just enqueues new errors in the same way as we capture analytic events and then flushes right away if needed by calling `flush`
so not another HTTP code path and no specific timeouts for this

src/error_tracking.rs
```diff
@@ -529,6 +592,147 @@ fn is_internal_capture_frame(function: &str) -> bool {
...
+    if let Some(location) = panic_info.location() {
+        event.insert_prop("$exception_panic_file", location.file())?;
+        event.insert_prop("$exception_panic_line", location.line())?;
+        event.insert_prop("$exception_panic_column", location.column())?;
+    }
+    // Reserved $exception_* properties are written after the location
+    // properties so they can't be overridden.
+    exception.write_into(&mut event, et_options)?;
+
+    send_panic_exception(options, event)
```
marandaneto, line 652: instead of doing this, we should call the posthog instance (global?) with `capture_exception(x)` and let the instance enqueue and flush everything
a big gap by doing this here is that we are bypassing eg `before_send`
we can always invoke the before send wrapped by catch_unwind to avoid the double panic
if the SDK isnt designed to do that, i think we should fix that first

cat-ph: yeah but I don't think catch_unwind helps unfortunately, I ~~tried~~ (asked the 🤖 to try) it:

```rs
use std::panic::{self, AssertUnwindSafe};

fn main() {
    let prev = panic::take_hook();
    panic::set_hook(Box::new(move |info| {
        // Simulate before_send / capture work that itself panics,
        // wrapped in catch_unwind
        let caught = panic::catch_unwind(AssertUnwindSafe(|| {
            println!("  [hook] running, about to panic inside hook...");
            panic!("panic raised INSIDE the hook (e.g. a before_send hook)");
        }));
        match caught {
            Ok(_) => println!("  [hook] catch_unwind returned Ok"),
            Err(_) => println!("  [hook] catch_unwind CAUGHT the inner panic -> safe"),
        }
        prev(info);
    }));

    println!("[main] triggering outer panic, caught by catch_unwind...");
    let outer = panic::catch_unwind(|| {
        panic!("outer panic");
    });
    println!("[main] outer catch_unwind returned: {:?} (if you see this, no abort)", outer.is_err());
}
```

=>

```
=== running (exit code is the verdict: 134 = SIGABRT) ===
[main] triggering outer panic, caught by catch_unwind...
  [hook] running, about to panic inside hook...
panicked at main.rs:10:13:

thread panicked while processing panic. aborting.
EXIT_CODE=134
```

what would help is running in in a _separate_ thread (like we do with the HTTP request) where the panic count is 0 🤔 

> if the SDK isnt designed to do that, i think we should fix that first

100% agree with this though, yeah

src/client/mod.rs
cat-ph: > should have a runtime-independent transport layer, where the panic just enqueues new errors

_true_, I kinda worked around not having this, I think our capture() has just a single event POST, no batching no queueing, no flush.. I think kinda similar to your point below, this was also to avoid the double panic and async capture, doing the manual request, but I don't particularly like it either 😬

src/error_tracking.rs
cat-ph: I could try and put the before_send on the separate thread to avoid bypassing it & get it working, but maybe the better question is, do we want to do that (I'd do that either way though), or do we want to try and refactor the SDK first to do the transport work?

src/client/mod.rs
marandaneto: yeah now that i understand the double panic issue, i understand the reasoning
but i'd push back on doing this right
once we add public APIs, we cant remove them and have to support moving forward

src/error_tracking.rs
marandaneto: i'd 100% go with the transport work since this is beneficial for the SDK anyway
happy to pair up/review if needed
you can try to point an agent to https://github.com/PostHog/sdk-specs and another SDK as an example (eg python?) and get something quite quick

marandaneto: since its 0.x, we can also do a breaking change if needed

cat-ph: thanks 💙 giving it a try then

src/error_tracking.rs
```diff
@@ -529,6 +592,147 @@ fn is_internal_capture_frame(function: &str) -> bool {
...
+    }
+
+    let et_options = options.error_tracking();
+    let exception = Exception::from_panic_info(panic_info, et_options.capture_stacktrace());
+
+    let mut event = Event::new_anon("$exception");
+    if let Some(location) = panic_info.location() {
+        event.insert_prop("$exception_panic_file", location.file())?;
+        event.insert_prop("$exception_panic_line", location.line())?;
+        event.insert_prop("$exception_panic_column", location.column())?;
```
ablaszkiewicz, line 646: not a Rust expert (at all) obviously but why do we have to append this as custom properties vs relying on just that `capture_stacktrace()` or adding `panic_info.location()` to the stack trace manually if capturing stack trace failed for whatever reason?

src/error_tracking.rs
cat-ph: put up a first version in https://github.com/PostHog/posthog-rs/pull/145 (a lot of code moving around 😅 ) & tested locally a bit, I'll stack / rebuild this one on top of that

marandaneto:
moving to draft until we get https://github.com/PostHog/posthog-rs/pull/145 so its out of our review filters

cat-ph:
closing in favor of https://github.com/PostHog/posthog-rs/pull/157 - I rewrote a lot of the logic there to use the new transport so just using a new one instead.

PR #131, opened 2026-06-09
Add custom user agent support

s6nedwards:
When using the rust client the Posthog api does not return server flags. This is because of 2 reasons:
- The rust client does not send up a user agent which causes the Posthog server treat requests as a user(public) request even if you include a personal api key(secure flag token)
- Even if sending up a user agent in the same style of other libraries because the Posthog server still does not recognize the rust client so it assumes the request is coming from a user(public) and not a server.

While this PR wont change behavior, at least until the posthog server adds rust client support, it allows end users to input a custom user agent to lie to the posthog server so they can retrieve server flags.

Manually and with automated testing. I added tests that explicitly check for user agent using the HTTP Mock library.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -85,6 +85,7 @@ impl BeforeSendHook {
         (hook)(event)
     }
 }
+const POSTHOG_RUST_USERAGENT: &str = "posthog-node/rust";
 
 /// Configuration options for the PostHog client.
 ///
@@ -204,6 +205,9 @@ pub struct ClientOptions {
     #[builder(setter(skip))]
     #[builder(default = "EndpointManager::new(DEFAULT_HOST.to_string())")]
     endpoint_manager: EndpointManager,
+
+    #[builder(default = "POSTHOG_RUST_USERAGENT.to_string()")]
+    user_agent: String,
 }
 
 /// Resolved client-level default properties for capture requests.
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -86,6 +86,11 @@ impl BeforeSendHook {
     }
 }
 pub const POSTHOG_RUST_USERAGENT: &str = "posthog-node/rust";
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+pub fn get_default_user_agent() -> String {
+    format!("{}/{}", POSTHOG_RUST_USERAGENT, VERSION)
+}
 
 /// Configuration options for the PostHog client.
 ///
@@ -206,7 +211,7 @@ pub struct ClientOptions {
     #[builder(default = "EndpointManager::new(DEFAULT_HOST.to_string())")]
     endpoint_manager: EndpointManager,
 
-    #[builder(default = "POSTHOG_RUST_USERAGENT.to_string()")]
+    #[builder(default = "get_default_user_agent()")]
     user_agent: String,
 }
 
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -85,11 +85,11 @@ impl BeforeSendHook {
         (hook)(event)
     }
 }
-pub const POSTHOG_RUST_USERAGENT: &str = "posthog-node/rust";
-const VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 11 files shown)


dustinbyrne:
cc @PostHog/team-feature-flags - seems like we'd prefer to fix this in the backend? What are your thoughts on this?

haacked:
@dustinbyrne yup!

Backend fix is up: PostHog/posthog#62594 adds `posthog-rust` to the feature-flags service's User-Agent allowlist, so `posthog-rust/<version>` will be treated as a server-side SDK and get `evaluation_runtime: "server"` flags back (and show up under its own name in per-SDK analytics instead of "other").

We still need the SDK side of this PR though: reqwest doesn't send any User-Agent by default, so without the default `posthog-rust/<version>` header the backend has nothing to recognize. What we don't need anymore is the custom user-agent override for impersonating another SDK; I'd trim this PR down to just sending the default header.

src/client/mod.rs
```diff
@@ -49,6 +49,13 @@ pub use async_client::client;
 #[cfg(feature = "async-client")]
 pub use async_client::Client;
 
+const VERSION: &str = env!("CARGO_PKG_VERSION");
```
marandaneto: lets use this const here as well https://github.com/PostHog/posthog-rs/blob/d1f1d3dd721a35d11dc0e31523ccf31e833fe4ad/src/event.rs#L260
so we dont dupe the CARGO_PKG_VERSION str

src/client/mod.rs
```diff
@@ -49,6 +49,13 @@ pub use async_client::client;
...
+const POSTHOG_RUST_USERAGENT: &str = "posthog-rust";
```
marandaneto: it shoudl be `posthog-rs`
and lets reuse the const here as well https://github.com/PostHog/posthog-rs/blob/d1f1d3dd721a35d11dc0e31523ccf31e833fe4ad/src/event.rs#L256

dustinbyrne (commented):
> We still need the SDK side of this PR though: reqwest doesn't send any User-Agent by default, so without the default `posthog-rust/<version>` header the backend has nothing to recognize. What we don't need anymore is the custom user-agent override for impersonating another SDK; I'd trim this PR down to just sending the default header.

Agreed, it'd be preferred that the header _not_ be configurable, as that would affect backend detection of the request as originating from a server-side SDK. @s6nedwards does this sound reasonable to you?

s6nedwards:
@dustinbyrne Yeah thats fine. I am not beholden to the ability to configure it. It was a stop gap that I had already implemented to get it to respond properly for my needs. I am unsure of the timeline to fixing the server so I left that in there in case someone else might want or need it. Fixing it on the server side is preferably with a fixed header to mirror the behavior of the other SDKs. Ill update this PR.

haacked:
> I am unsure of the timeline to fixing the server

I just merged https://github.com/PostHog/posthog/pull/62594, so it should be deployed by the time we get this merged and deployed.

marandaneto:
@s6nedwards mind adressing the comments so we can merge/release this? thanks

src/client/mod.rs
s6nedwards: Its `posthog-rust` to match the other SDKs which use the format `posthog-{lang/platform}`  and that matches the new server value check.

s6nedwards:
@marandaneto Ive updated the PR to address the comments. Ive replied to the one comment I didnt address.

src/client/mod.rs
marandaneto: Nope, the backend changed to -rs since its how we report everything already.
This should match what we send as $lib as well

marandaneto:
> @marandaneto Ive updated the PR to address the comments. Ive replied to the one comment I didnt address.

thanks, replied, still one left!

s6nedwards:
@marandaneto updated to be `posthog-rs`

marandaneto:
superseeded by https://github.com/PostHog/posthog-rs/pull/143 so we close the loop faster here

PR #132, opened 2026-06-09
fix(capture): route flag events through capture + papercuts

eli-r-ph:
## Summary
Follow-on that needed to land prior to un-gating capture v1, and misc. papercuts identified in smoke testing.

- **`$feature_flag_called` routes through v1** when `capture-v1` is enabled
    - This previously went around the std SDK capture path
    - v1 support stays fire-and-forget with a **single attempt and no retry loop** on both async and blocking hosts
- **Accept any 2xx as success** in `v1_capture::after_response` (was exactly 200), so a future 201/202/207 from capture is not misclassified as a connection error. Malformed bodies on 2xx still surface as `Error::Serialization`.
- **Canonical SDK identity**: `posthog-sdk-info` / `user-agent` now send `posthog-rs/<version>` (was `posthog-rust/<version>`)
    - The name segment now matches the `$lib` value the v0 path sends, so  v1 backend attributes correctly in SDK Health, usage reports, and Library columns
- **CI**:  ensure full matrix of v1-relevant Rust build features are exercised in CI unit test runs

## Test plan

- [x] `cargo test` (async, v0): 183 pass
- [x] `cargo test --features capture-v1`: 226 pass
- [x] `cargo test --no-default-features` (blocking, v0): 185 pass
- [x] `cargo test --no-default-features --features capture-v1`: 225 pass
- [x] `cargo test --workspace`, `cargo fmt --check`, `cargo clippy -- -D warnings`
- [x] New tests: 2xx-range unit + e2e (201), sdk-info format pin (unit + e2e), `build_flag_event_request` (batch shape, defaults, compression), v1 flag-event routing + single-attempt-failure (async + blocking)

### DIFF
--- a/.sampo/changesets/capture-v1-2xx-sdk-info-flag-events.md
+++ b/.sampo/changesets/capture-v1-2xx-sdk-info-flag-events.md
@@ -1,0 +1,9 @@
+---
+cargo/posthog-rs: patch
+---
+
+Three fixes to the unstable `capture-v1` pipeline (off by default):
+
+- Accept any 2xx HTTP status as success on V1 capture responses instead of exactly 200, so a future 201/202/207 from capture is not misclassified as a connection error. Malformed bodies on 2xx still surface as `Error::Serialization`.
+- Send the canonical SDK identity `posthog-rs/<version>` (previously `posthog-rust/<version>`) in the `posthog-sdk-info` and `user-agent` headers. The name segment now matches the `$lib` value the v0 path sends, so capture-side `$lib`/`$lib_version` materialization attributes V1 traffic correctly in SDK Health, usage reports, and Library columns.
+- Route `$feature_flag_called` events through the V1 analytics endpoint when `capture-v1` is enabled (previously they always took the legacy v0 path, splitting the pipeline). Shipping stays fire-and-forget with a single attempt and no retry loop, matching the v0 flag-event semantics on both the async and blocking clients.
--- a/src/client/v1_capture.rs
+++ b/src/client/v1_capture.rs
@@ -45,12 +45,8 @@ pub(crate) fn build_events(events: &[Event], defaults: &CaptureDefaults) -> Vec<
 
 pub(crate) fn build_headers(opts: &ClientOptions, request_id: &Uuid, attempt: u32) -> HeaderMap {
     let version = env!("CARGO_PKG_VERSION");
-    // V1 SDK-identity standard: `<canonical-$lib-name>/<semver>`. The name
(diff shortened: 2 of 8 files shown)


src/client/v1_capture.rs
```diff
@@ -584,4 +620,149 @@ mod tests {
...
+                &mut pending,
+                &mut final_results,
+            );
+            assert!(
+                matches!(step, Step::Done),
+                "HTTP {status} should be treated as success"
+            );
+            assert_eq!(final_results.len(), 1);
+        }
+    }
```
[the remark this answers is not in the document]
eli-r-ph: Repo style is loop-over-cases for status-code sweeps (see `src/client/retry.rs`, `tests/test_v0_retry.rs`, `tests/test_v1_capture.rs`, `tests/test_blocking.rs`) and the crate has no rstest dependency. The assert message includes the failing status, so the regressed case is identified in the panic output. Keeping the loop.

src/client/v1_capture.rs
```diff
@@ -183,7 +206,9 @@ pub(crate) fn after_response(
     pending: &mut Vec<V1Event>,
     final_results: &mut HashMap<Uuid, EventResult>,
 ) -> Step {
-    if status == 200 {
+    // The backend sends exactly 200 today; accept the whole 2xx class so a
+    // future 201/202 isn't misclassified as a connection error.
+    if (200..=299).contains(&status) {
         let batch_resp: CaptureResponse = match serde_json::from_str(body) {
```
ioannisj, line 212: Empty body would err here right? but there are some 2xx that traditionally have empty body. Probably not our case since from the comments I see we care about specific 2xx codes

ioannisj (approved):
LG with my very limited rs skills

.sampo/changesets/capture-v1-2xx-sdk-info-flag-events.md
```diff
@@ -0,0 +1,9 @@
+---
```
jose-sequeira, line 1: Is this intentional?

jose-sequeira (approved):

src/client/v1_capture.rs
eli-r-ph: This should be fine atm, and sendBeacon (etc.) 204 response support isn't needed in this SDK...but as I think about it more, I want to check on another potential backend/frontend interaction corner case before I land this.

Thanks for mentioning it! 🙇

eli-r-ph: Updated to shore up the empty body case and add a test for it in the v1 capture_batch codepath. Thanks again!

PR #134, opened 2026-06-10
feat: capture instruction addresses and debug images for symbolication

cat-ph:

Production Rust builds strip debug info, so client-side `backtrace` resolution yields empty or mangled frames. This makes the SDK emit what PostHog's server-side native symbolication needs (mirroring the iOS SDK contract): raw `instruction_addr`/`symbol_addr`/`image_addr` per frame under `platform: "native"`, plus an event-level `$debug_images` list (debug id, load address, size, code file, arch) built from the modules loaded in the process. Debug symbols are uploaded separately via `posthog-cli debug-symbols upload` (ELF) or `dsym upload` (macOS).

Client-side enrichment (`function`/`filename`/`lineno`) still rides along as the fallback for un-uploaded symbols. Second of two stacked PRs (on the frame-order fix).

Key mechanics:

- Module enumeration via `findshlibs` (one pass per capture, no file IO): GNU build ids → debug ids using the same little-endian GUID interpretation as `symbolic` on the upload side (vector-pinned in tests, deliberately not gated on target endianness since the upload pipeline computes ids on little-endian hosts); Mach-O `LC_UUID` uppercased to match `dsym upload` chunk ids; PDB GUID+age handled for the future Windows path.
- Address-only frames (stripped builds) default `in_app` by module: main executable yes, shared libraries no; named frames keep the existing path/function rules.
- SDK-frame stripping no longer depends on symbol names: capture flows through a monomorphic `#[inline(never)]` chokepoint (`Exception::from_items_captured`) whose entry address is compared against frame symbol addresses; the frame directly above it (constructor or client wrapper, both generic) is dropped with it. Documented caveat: fully stripped Apple/Windows builds can't resolve entry addresses (no `_Unwind_FindEnclosingFunction`), so the SDK prefix survives there as address-only frames — within the agreed v1 platform scope (Linux is the fully-supported target).
- `$debug_images` is filtered to images referenced by frames that survive `max_frames` trimming.
- `StackFrame` drops the `resolved` field and gains the address fields — a breaking shape change to a struct only exposed on the unreleased `error-tracking` feature branch; no published version carries it.

Agent-authored (Claude Code, directed by @cat-ph); automated tests only:

- New tests: build-id → debug-id conversion vectors (verified against `symbolic`'s output for the cymbal test fixtures), module range matching, and a self-referential capture test asserting every frame has a parseable address, `$debug_images` is present with the right `type`/`arch`, and every image is referenced by a frame.
- Full gate on both feature sets: `cargo fmt --check`, `cargo check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (202 + 204 passed).
- Three second-model review rounds drove real fixes (PDB `debug_id()` accessor + GUID byte-swap, symbol-entry vs ip addresses, wrapper-frame stripping); the two remaining findings were rejected with rationale documented in code comments (Apple/Windows stripped-build stripping is v1-out-of-scope; endianness gating would mismatch our little-endian upload pipeline).

## :pencil: Checklist

### If releasing new changes

- [ ] Ran `sampo add` to generate a changeset file — covered by the branch's existing `rust-error-tracking` changeset; extends unreleased behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -808,6 +808,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
 
 [[package]]
+name = "findshlibs"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "40b9e59cd0f7e0806cca4be089683ecb6434e602038df21fe6bf6711b2f07f64"
+dependencies = [
+ "cc",
+ "lazy_static",
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "fixedbitset"
 version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1781,6 +1793,7 @@ dependencies = [
  "ctor",
  "derive_builder",
  "dotenv",
+ "findshlibs",
  "flate2",
  "futures",
  "httpmock",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ derive_builder = "0.20.2"
 uuid = { version = "1.13.2", features = ["serde", "v7"] }
 # Pinned: backtrace 0.3.75+ raises its MSRV to 1.82, above this crate's 1.78 (see rust-version).
 backtrace = { version = "=0.3.74", optional = true }
+findshlibs = { version = "0.10", optional = true }
 sha1 = "0.10"
 regex = "1.10"
 tracing = "0.1"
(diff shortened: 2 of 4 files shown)


cat-ph:
Restacked on the updated frame-order branch (now on top of the panic PR). Integration changes while rebasing: panic captures route through the shared from_items_captured chokepoint (panics now get instruction addresses and $debug_images too, and the panic-machinery frame names folded into the shared strip list); every direct chokepoint caller is #[inline(never)] so the caller-frame drop can never eat an application frame under release inlining; native address fields are omitted for frames whose module has no usable debug id (clean client-resolved pass-through instead of a guaranteed server-side resolution failure); debug image arch reports arm64-style names; Exception::from_error was dropped (unused since build_exception_event pairs error_items with the chokepoint directly).

PR #145, opened 2026-06-15
feat: runtime-independent background event transport

cat-ph:
## What

Replaces the per-`capture` direct HTTP POST with a runtime-independent background transport: a single `std::thread` worker draining a bounded channel with **blocking** reqwest (never a tokio task), so it works for the async client, the blocking client, and a future panic hook with no runtime present.

- `capture()` / `capture_batch()` become **non-blocking enqueues**.
- New: `flush()` (awaited on the async client, blocking on the blocking client), `shutdown()` (flush + stop + join; idempotent; drops further captures), flush-on-`Drop`, `pending_events()`, and global `flush()` / `shutdown()`.
- Batching knobs: `flush_at`, `max_batch_size`, `flush_interval_ms`; a bounded `max_queue_size` queue that drops with a single warning when full.
- Worker retry: keep-on-transient (5xx/408/429/network) with exponential backoff honoring `Retry-After`, drop-on-terminal — reusing the existing sans-IO `retry.rs` / `v1_capture` decision logic.
- `before_send` now runs on the worker, so it applies to every queued event (including future panic captures).
- v0 single capture now batches through `/batch/`.
- An injectable `Clock` (`now` + `now_utc`) makes interval-flush, retry backoff, and v1 wire timestamps deterministic in tests — a `ManualClock` plus a test-only `Tick` drive virtual time, with no real sleeps.
- The compliance adapter delegates batching/flush/shutdown to the SDK (adds a `/shutdown` route).

## Breaking change (0.x)

`capture()` / `capture_batch()` return `Ok(())` as soon as the event is queued instead of awaiting delivery, and HTTP failures surface as **logged warnings**, not `Err`. **Call `flush()` or `shutdown()` before process exit** to ensure queued events are delivered. The global client lives in a `static`, whose `Drop` never runs, so global users must call `posthog_rs::shutdown()` / `posthog_rs::flush()` explicitly.

## Specs

Implements the `event-batcher`, `retry-queue`, `http-client`, `flush`, and `shutdown` contracts. The flush/shutdown acceptance scenarios run through the compliance adapter; the internal-behavior specs are covered by SDK tests (`tests/test_transport.rs` plus virtual-clock unit tests in `transport.rs`).

## Out of scope / deferred

- **Panic autocapture** is handled separately (`main` has no panic hook to migrate); the worker already exposes the blocking flush a future panic path needs.
- `413` batch-shrinking (spec-optional) is not implemented, matching prior behavior.

## Validation

**TL;DR:** green across all four feature configs (fmt, clippy, unit + integration tests), and verified end-to-end against a local PostHog dev stack — events actually ingest on both clients and both wire versions.

- `cargo fmt --check` clean; `cargo clippy --all-targets` clean (async v0 + v1).
- Tests green across all four feature configs: async-v0 229, blocking-v0 220, async-v1 274, blocking-v1 260.
- `api/public-api.txt` regenerated with the pinned nightly (+11 lines: `flush` / `shutdown` / `pending_events` / `Drop` + four builder setters + global `flush` / `shutdown`).

<details><summary>Local end-to-end run against a dev PostHog stack</summary>

A throwaway harness (`capture` ×5 with a small `flush_at`, then `capture_exception`, a final `capture`, and `shutdown`) ran on each transport config; every event was tagged with a unique marker and queried back from ClickHouse. All ingested — 7/7 per config (5 capture + 1 shutdown-drain + 1 exception):

| Config | Ingested |
|---|---|
| async + v0 | 7 / 7 |
| async + v1 | 7 / 7 |
| blocking + v0 | 7 / 7 |

Confirms end-to-end:

- the `std::thread` + blocking-reqwest worker delivers both from inside a tokio runtime (async client) and from a blocking `main`;
- both `/batch/` (v0) and `/i/v1/analytics/events` (v1) ingest, with v1 returning per-event `ok`;
- batching split as designed — a threshold flush, an explicit `flush()`, and a `shutdown()` drain;
- `flush()` and `shutdown()` both drain (the event captured just before `shutdown()` landed);
- the manual exception path ingests;
- `pending_events()` read `0` after both flush and shutdown.

`blocking + v1` was not run e2e (every dimension is covered by the other three) but passes the full test suite.

</details>


src/client/transport.rs
```diff
@@ -0,0 +1,904 @@
...
+pub(crate) struct TransportHandle {
+    tx: mpsc::Sender<Control>,
+    /// Pending `Capture` events not yet pulled by the worker. Gates the bounded queue.
+    len: Arc<AtomicUsize>,
+    /// Set once `shutdown`/`Drop` begins; blocks further enqueue and control sends.
+    closed: AtomicBool,
+    worker: Mutex<Option<JoinHandle<()>>>,
+    /// Latches the single "queue full" warning so a full queue doesn't spam logs.
+    full_warned: AtomicBool,
+    max_queue_size: usize,
```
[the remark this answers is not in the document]
cat-ph: Fixed in 217fb0f — `full_warned` now re-arms once the queue drains back to empty, so it warns once per full episode instead of once per process lifetime. Thanks!

compliance/adapter/src/main.rs
```diff
@@ -308,23 +262,23 @@ async fn flush(
...
                     .into_response();
             }
         }
-    }
+    };
 
-    let flushed = flush_buffer(&state.instances, &key).await;
+    let before = client.pending_events() as u64;
+    client.flush().await;
+    let flushed = before.saturating_sub(client.pending_events() as u64);
```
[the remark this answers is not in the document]
cat-ph: Good catch — fixed in 217fb0f: `pending_events()` now counts everything in flight (channel + worker buffer + retries), not just channel depth, so this before/after delta is accurate.

compliance/adapter/src/main.rs
```diff
@@ -333,20 +287,45 @@ async fn flush(
...
+                total_events_captured: s.total_events_captured,
+                total_events_sent: (s.total_events_captured as i64 - pending).max(0) as u64,
+                total_retries: 0,
+                last_error: s.last_error.clone(),
+            }))
+            .into_response()
+        }
         None => Json(serde_json::json!(StateResponse {
             pending_events: 0,
             total_events_captured: 0,
```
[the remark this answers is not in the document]
cat-ph: Acknowledged — this only over-counts if the SDK queue actually fills, which the acceptance suite never does. An exact sent-count would need an SDK delivery callback; left as a documented follow-up (see the note now in `get_state`).

src/client/mod.rs
```diff
@@ -190,6 +191,25 @@ pub struct ClientOptions {
...
+
+    /// Interval between automatic time-based flushes, in milliseconds
+    /// (default: 5000).
+    #[builder(default = "5000")]
+    pub(crate) flush_interval_ms: u64,
+
+    /// Maximum number of events buffered before new events are dropped
+    /// (default: 10000). A single warning is logged while the queue is full.
+    #[builder(default = "10000")]
+    pub(crate) max_queue_size: usize,
```
cat-ph, line 211: those defaults are ~~a kind of mix between py and go~~ now pretty much the same with py

api/public-api.txt
```diff
@@ -107,10 +107,15 @@ pub async fn posthog_rs::Client::capture_exception<E>(&self, &E) -> core::result
...
 pub async fn posthog_rs::Client::is_feature_enabled<K: core::convert::Into<alloc::string::String>, D: core::convert::Into<alloc::string::String>>(&self, K, D, core::option::Option<std::collections::hash::map::HashMap<alloc::string::String, alloc::string::String>>, core::option::Option<std::collections::hash::map::HashMap<alloc::string::String, serde_json::value::Value>>, core::option::Option<std::collections::hash::map::HashMap<alloc::string::String, std::collections::hash::map::HashMap<alloc::string::String, serde_json::value::Value>>>) -> core::result::Result<bool, posthog_rs::Error>
+pub fn posthog_rs::Client::pending_events(&self) -> usize
+pub async fn posthog_rs::Client::shutdown(&self)
+impl core::ops::drop::Drop for posthog_rs::Client
+pub fn posthog_rs::Client::drop(&mut self)
```
marandaneto, line 118: i think pending_events, and drop shouldn't be public APIs

src/client/mod.rs
marandaneto, line 211: lets match them with python if not yet

marandaneto:
maybe ask a Rust expert to take a look, as in someone who often contributes Rust code, e.g., Hughes, Ingestion folks, etc.?

cat-ph:
thanks! yep, pinging @eli-r-ph too 🙏  since I think he's also a bit familiar with the logic here (and I'm also changing the capture v1 stuff 😁)

api/public-api.txt
cat-ph: I think drop will end up being because we do flush on drop (it's technically not user callable) and we do the same for the FlagPoller / AsyncFlagPoller

but agree for pending_events, I think I can gate it with feature = "test-harness" but it probably still shows up here, checking 👀

src/client/mod.rs
cat-ph: they are, I think the only difference here is 5s flush interval vs 0.5s in python - it's 5s as in go and I think that might be a bit simpler / less requests at lower volume, WDYT?

marandaneto: yeah 0.5s is surprising tbh, very low, but it really depends a lot on the application, eg 10k of users/minutes the lower the better, but if you dont have lots of users, higher the better

api/public-api.txt
marandaneto: i think the test harness should be able to wait for a certain amount of events or time so probably not necessary but i didnt look that closely
the public api snapshot wont consider testing features since the generator does not consider that, but if its only used for the testing harness and under a testing feature, i think thats ok, just make sure that the feature is test-only/internal in the code blocks

src/client/mod.rs
cat-ph: indeed, there's still `flush_at` in the mix which should "dominate" here at high volume, in this case the flush interval is a backstop for the low-volume side

api/public-api.txt
cat-ph: yeah it's a required contract field in the sdk test harness, the compliance adapter depends on it 😬 I've put it under a testing feature so it's internal only

eli-r-ph:
👋 hey @cat-ph thanks for doing this! it's something I was hoping to circle back to soon myself ❤️ couple things to look into I think:

Async capture path: I don't think having a single thread consumer (no pool) is safe here, any retried or higher-latency request will cause head-of-line blocking and could fill up the queue. I'd suggest using `tokio::spawn`  (or runtime with a thread pool no bigger than `num_cpus` or  smaller than `(num_cpus / 2) + 1`, if you're worried about starving the client app) to avoid this.

drop/flush/shutdown handling: I don't think this should block for very long as it could hold up app shutdown, especially if the queue is full or the active (head-of-line-blocked) event publish is busy. I think flush should be async and fire-and-forget for safety. 

The core idea to drop new events enqueued when the queue is full is sound 👍 

Capture timestamp handling:
* capture v0:
    * `event.timestamp` if unset by the caller should be `now` at _capture (enqueue) time_
    * `event.sent_at` should be set at publish time
* capture v1:
   * `event.timestamp` if unset by caller should be `now` at capture (enqueue) time also
   * Timestamp header and batch `created_at` should be set at publish time
   
Other than that, this LGTM 👍

cat-ph:
thanks for looking @eli-r-ph 💙 

great catch with the timestamps, fixing it to be at capture time (& adding sent_at), the header/created_at were OK here but the timestamp wasn't 😬

I didn't go with tokio here to be able to use this in all 3 places that don't really share a runtime: the async client (w/ tokio), sync client (no async runtime) + std::panic in particular (arbitrary thread where runtime might not be present), I modeled it after go/py where both use a bg thread/goroutine

good point about the HOL block, I think for retries it's safe cause we defer it in the `retries: VecDeque<RetryBatch>` but the HTTP attempt itself still blocks, I think still worth doing a std-thread pool for this? might complicate ordering but I don't think we care since the timestamps are proper, just need a mutex on the retries VecDeque probably, taking a look! 👀

for the drop/flush/shutdown handling I'm not very sure, I'm definitely adding a shutdown timeout (good point), but I think flush should still be blocking? 🤔 since capture() is fire-and-forget and flush here was the only place to signal it's flushed (and it also returns on a failure, since that's rescheduled/retried separately) so not unbounded but blocking by design

src/client/async_client.rs
hpouillot: If we start introducing breaking changes, maybe we could update the signature. It doesn't need to be async and it always returns Ok, so maybe we could just have:
```
pub fn capture_batch(...) -> ()
```

src/client/async_client.rs
```diff
@@ -292,26 +296,49 @@ impl Client {
     /// # Remarks
     ///
     /// Disabled clients skip the request and return `Ok(())`.
+    /// Enqueue `event` for delivery. Non-blocking: the event is handed to the
+    /// background worker, which batches, sends, and retries it. Returns once the
+    /// event is queued — not once it is delivered. Disabled clients and a full
+    /// queue drop the event (the latter with a single warning).
     #[instrument(skip(self, event), level = "debug")]
     pub async fn capture(&self, event: Event) -> Result<(), Error> {
```
hpouillot: same here

src/client/transport.rs
```diff
@@ -0,0 +1,967 @@
...
+use super::ClientOptions;
+use crate::Event;
+
+/// Messages sent from producers (`capture`/`flush`/`shutdown`) to the worker.
+pub(crate) enum Control {
+    // Boxed so the channel node isn't sized to a whole Event on every flush/
+    // shutdown message (clippy::large_enum_variant).
+    Capture {
+        event: Box<Event>,
+        historical_migration: bool,
```
hpouillot: it feels overkill to have this on every event, do we really need it ? If I understand correctly, the SDK could be used for backfills, but then we could have something at setup time or another client right ?

src/client/async_client.rs
eli-r-ph: I think the idea is, when there's an actionable whole-request fail the end user wants to know about it and decide how to debug (they are doing something wrong in the typical 4xx case) so this let's them decide how to log/observe those. I don't think we should get rid of it unless we want to invest further in surfacing errors internally for them.

In capture v1 these errors get much more granular/actionable and will be returned more frequently than in legacy capture so the value will increase over time for folks debugging on the client side

src/client/blocking.rs
```diff
@@ -1107,3 +954,21 @@ impl Client {
...
+        };
+        if !transport.begin_close() {
+            return;
+        }
+        let (tx, rx) = std::sync::mpsc::channel();
+        if transport.send_control(Control::Shutdown(Completion::Blocking(tx))) {
+            let _ = rx.recv();
+        }
+        transport.join();
+    }
```
eli-r-ph, line 971: Can't this also block a long time in busy/fail/shutdown scenarios, or if the end user doesn't set up `shutdown()` correctly? 🤔

eli-r-ph:
@cat-ph yeah totally understand if introducing a runtime choice here may cause trouble for end users. Introducing some form of bounded thread pool is a good idea to avoid HOL. Better batching for capture submissions this PR introduces will amplify the risk. Something to think about. Ensuring the Drop behavior doesn't potentially cause trouble is worth a 2nd look too I think.

Thanks again for digging into this, the change will be useful! ❤️

src/client/blocking.rs
cat-ph: hmmmm, it _does_ but up to the shutdown timeout in [`968cca2` (commit in this PR)](https://github.com/PostHog/posthog-rs/pull/145/changes/968cca27db1c5e0fda8173faf72f3c9a20867c2b) I think (well technically up to that + the POST reqwest timeout which is another 30s), I have a draft in https://github.com/PostHog/posthog-rs/pull/147 stacked on this one to convert to a std-thread pool (but currently a bit more complex than I'd like, trying to simplify / test more thoroughly!)

src/client/async_client.rs
cat-ph: agree that Result doesn't really do much here *now* - it was kinda forward-compat/forward-looking to surfacing some errors back and not having to do another breaking change to revert it

however for actual requests failing, we'd need to add an `on_error` callback/channel to surface them, since this now just enqueues.

I took a look again through the py/go implementations, and I think there are two things:

1. we can keep this shape here, and return some "simple" enqueue validation errors (blank name, distinct id etc), queuefull, closed (that we currently just swallow)
2. adding an on_error callback for the async send errors (can also stack/follow-up this)

but I _think_ we can drop the async part since the validations still keep it synchronous

hpouillot: imo the default capture should be best effort to avoid discarding errors on every capture call. We could maybe have a `try_capture` that returns a Result to handle malformed events and a full queue if it is something user care about ?

cat-ph: true, tbf it makes sense to go with the intent then, since we expect it to be fire-and-forget. updating! and then we can add try_capture later if people find that useful

src/client/async_client.rs
dustinbyrne: this documentation is now stale, is this right?

src/client/transport.rs
```diff
@@ -0,0 +1,1096 @@
...
+            url_base,
+            clock,
+            len,
+            retries: VecDeque::new(),
+        }
+    }
+
+    fn send_batch(&mut self, events: Vec<Event>, historical_migration: bool, final_attempt: bool) {
+        let defaults = self.options.capture_defaults();
+        let count = events.len();
```
dustinbyrne, line 816: this looks like it can become stale after `before_send` filtering in `build_batch_payload`

src/client/transport.rs
cat-ph: yep, updated it to go separately, seems clearer/better!

src/client/blocking.rs
cat-ph: updated to propagate the deadline too to the reqwest timeout so it should be pretty safe now. (except for the whole std-thread pool which I'll stack in #147 - we can either ship separately or just review that one when it's ready separately and collapse into this so we "release" at the same time, but I'd defer it tbh)

src/client/transport.rs
cat-ph: good catch, ty, fixed!

src/client/async_client.rs
cat-ph: yep, updated it already but didn't push, was still testing it, fixed now!

cat-ph:
^ bunch of updates above, but mostly kept in separate commits, tl;dr:

- historical migration goes through separate path to simplify (Hugues' suggestion)
- fixed event count after `before_send` (Dustin's catch)
- a few small clamps for durations
- the shutdown deadline should be properly handled now (passed through reqwest etc, after discussion with Eli above)
- dropped async/result from the capture public API - should be simpler now since it's fire-and-forget (technically a breaking change 😅)

hpouillot (approved):
🦀 🦀 🦀

marandaneto (approved):

PR #155, opened 2026-06-24
fix: saturate transport in-flight counter to prevent underflow wedge

eli-r-ph:
The in-flight event counter is a shared `AtomicUsize` decremented from many paths, and that accounting is tricky. This hardens it against a double-decrement bug underflowing the counter.

## :bulb: Why

- An underflow wraps to a huge value.
  - `try_reserve` then sees the queue as permanently full.
  - Every later event is silently dropped for the life of the client.
- The counter is decremented from 5+ paths, so a future bug is plausible.
  - `before_send` drops, partial v1 results, terminal sends, shutdown drops, channel drain.
- Flagged in review on #145.

## :hammer: What

- Decrements now saturate at 0 instead of `fetch_sub`.
  - `fetch_update` + `saturating_sub` in `dec_len`.
- `debug_assert` catches a real underflow bug in tests.
  - Release builds clamp instead of wedging.
- Both "worker gone" enqueue paths now route through `dec_len` too.

## :green_heart: How tested

- `fmt` + `clippy -D warnings` clean (default + `capture-v1`).
- 14 `transport::` tests pass across async-v1, blocking-v1, v0.
  - Includes the two that drive `dec_len` to 0 with the assert live.
- No saturation-only test: it needs `--release` (assert panics in debug), so it would not run in CI.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.sampo/changesets/harden-in-flight-counter.md
+++ b/.sampo/changesets/harden-in-flight-counter.md
@@ -1,0 +1,5 @@
+---
+posthog-rs: patch
+---
+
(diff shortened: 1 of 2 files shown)


cat-ph (approved):
ooh, good point, thanks! 💙

Issue #151, opened 2026-06-19
Add request-scoped context and web middleware helpers


marandaneto:
## Summary

Add request-scoped identity/context support to the Rust SDK, plus middleware/tracing-header helpers for common Rust web frameworks.

## Motivation

Python has contexts (`new_context`, `identify_context`, `set_context_session`, `tag`) and Django middleware that can apply identity/session metadata to all events captured during a request. Go has request-context helpers and middleware. Rust currently requires callers to pass `distinct_id`, `$session_id`, and shared properties manually for each capture, exception, and feature flag evaluation.

This makes it harder to connect backend Rust events/errors/flag evaluations to frontend sessions and replays, especially when using PostHog JS tracing headers:

- `X-PostHog-Distinct-Id`
- `X-PostHog-Session-Id`

## Proposed scope

Provide a Rust-native way to set request-scoped PostHog context and framework integrations that populate it from incoming requests.

Possible API shape:

```rust
let ctx = PostHogContext::new()
    .distinct_id("user_123")
    .session_id("session_uuid")
    .property("request_id", request_id)?;

client.capture_with_context(ctx.clone(), event).await?;
let flags = client.evaluate_flags_with_context(ctx.clone(), options).await?;
client.capture_exception_with_context(&error, ctx).await?;
```

Or, if there's a more idiomatic Rust approach, use task-local/request-extension state and helpers.

Middleware candidates:

- `tower`/`axum` first, since this covers a broad set of Rust web apps.
- Actix support can be part of this issue or a follow-up depending on implementation size.

Middleware should be able to:

- Extract `X-PostHog-Distinct-Id` and `X-PostHog-Session-Id` from incoming requests.
- Attach request-scoped defaults for captures and exceptions.
- Make the request-scoped distinct ID available for feature flag evaluation.
- Allow an authenticated server-side distinct ID to override client-controlled tracing headers for security-sensitive events/decisions.

## Acceptance criteria

- Rust SDK has a request-scoped context API for distinct ID, session ID, and shared properties/tags.
- Captures, exceptions, and feature flag evaluations can consume this context.
- At least one middleware integration (preferably `tower`/`axum`) extracts PostHog tracing headers and populates context.
- Docs clearly warn that tracing headers are client-controlled analytics context, not authentication/authorization.
- Tests cover context propagation and header extraction.

cat-ph:
[...]

PR #157, opened 2026-06-25
feat: panic autocapture

cat-ph:
## What

Adds panic autocapture to the Rust SDK, rebuilt on top of the now-merged runtime-independent transport.

When the global client is initialized with `init_global`, a process-wide `std::panic` hook is installed **by default** — crash reporting, on by default. It captures each panic as a personless `$exception` event (the panic payload, panic-site `$exception_panic_file`/`_line`/`_column`, and a call-site stack trace honoring `capture_stacktrace`), then chains the previously installed hook.

Capture routes through the transport's background worker, so:
- no async runtime is required (works for both the async and blocking clients);
- `before_send` hooks still apply; and
- a panic inside `before_send` is a joinable worker-thread panic, not a double-panic process abort.

## Public API

- `ErrorTrackingOptions::capture_panics` (default `true`) — set `false` to opt out of the automatic global install.
- `ErrorTrackingOptions::panic_flush_timeout_ms` (default `2000`) — how long the hook blocks the panicking thread waiting for the flush before letting the panic proceed. Kept short and separate from `shutdown_timeout_ms` so a slow/unreachable endpoint can't freeze the crash or delay its panic message.
- `install_panic_hook(Arc<Client>)` — install the hook for a *standalone* (non-global) client. It takes an `Arc` because the hook is `'static`; a disabled client is a no-op and never latches the single process-wide hook.
- New error variant `Error::PanicHookAlreadyInstalled`.

## Notes

- Delivery is best-effort: the `$exception` is flushed ahead of queued retries, but under sustained transport backpressure (slow/unreachable endpoint or a large capture backlog) it may not be sent within the bound before the process exits. A true queue-jumping priority path is better suited to the sender-pool rework than this std-mpsc worker.
- On the panicking thread the hook does only panic-safe, tracing-free work (enqueue + bounded flush); the HTTP send and `before_send` run on the worker.

## Testing

`cargo test` across the four feature configs (default, `capture-v1`, `--no-default-features --features error-tracking` ± `capture-v1`); `clippy -D warnings`; `cargo doc`; public-api snapshot regenerated.

tested on a devbox

CleanShot 2026-06-25 at 19 56 46


marandaneto:
`$exception_level` should be `fatal` for panics

.sampo/changesets/panic-autocapture.md
```diff
@@ -0,0 +1,11 @@
+---
+posthog-rs: minor
+---
+
+Panic autocapture, on by default. When the global client is initialized with `init_global`, a process-wide `std::panic` hook is installed automatically (gated by the default-on `error-tracking` feature). It captures panics as personless `$exception` events through the global client, then calls the previously installed hook. Each event carries the panic payload, the panic-site location (`$exception_panic_file`/`_line`/`_column`), and a call-site stack trace honoring `capture_stacktrace`. Capture routes through the background worker thread, so it needs no async runtime and a panicking `before_send` hook can't abort the process.
```
marandaneto: i think we need to find a way also to allow capturing an identified person exception instead of only personless, not a blocker, but a possible improvement
i think python has something similar with 

```
with posthog.new_context()
  // identify
  do things within this context
  // capture errors
```

api/public-api.txt
```diff
@@ -178,9 +179,11 @@ pub fn posthog_rs::ErrorTrackingOptions::default() -> Self
 pub struct posthog_rs::ErrorTrackingOptionsBuilder
 impl posthog_rs::ErrorTrackingOptionsBuilder
 pub fn posthog_rs::ErrorTrackingOptionsBuilder::build(&self) -> core::result::Result<posthog_rs::ErrorTrackingOptions, posthog_rs::ErrorTrackingOptionsBuilderError>
+pub fn posthog_rs::ErrorTrackingOptionsBuilder::capture_panics(&mut self, bool) -> &mut Self
 pub fn posthog_rs::ErrorTrackingOptionsBuilder::capture_stacktrace(&mut self, bool) -> &mut Self
 pub fn posthog_rs::ErrorTrackingOptionsBuilder::in_app_exclude_paths(&mut self, alloc::vec::Vec<alloc::string::String>) -> &mut Self
 pub fn posthog_rs::ErrorTrackingOptionsBuilder::in_app_include_paths(&mut self, alloc::vec::Vec<alloc::string::String>) -> &mut Self
+pub fn posthog_rs::ErrorTrackingOptionsBuilder::panic_flush_timeout_ms(&mut self, u64) -> &mut Self
```
marandaneto: i have the feeling we dont need this
The SDK should have a flush method
the panic hook should just call posthog.capture_panic(x), and then `posthog.flush()`
the SDK config already has a flush config, why do we need a special one for panic?
i am ok having a special panic flush timeout, just need to understand why

src/client/async_client.rs
```diff
@@ -316,6 +316,60 @@ impl Client {
         }
     }
 
+    /// Whether the client is disabled (no transport; capture is a no-op). Used
+    /// by the panic hook to skip building an event it could never send.
+    #[cfg(feature = "error-tracking")]
```
marandaneto, line 321: we probably dont need this
`capture_panic` should have access to this private method and bail out if disabled, we dont need to expose as a public method unless there's reasons/limitations

src/client/async_client.rs
```diff
@@ -316,6 +316,60 @@ impl Client {
...
+    pub(crate) fn is_disabled(&self) -> bool {
+        self.options.is_disabled()
+    }
+
+    /// The client's Error Tracking options, used by the panic hook to build
+    /// panic exception events with the client's configured policy.
+    #[cfg(feature = "error-tracking")]
+    pub(crate) fn error_tracking_options(&self) -> &crate::error_tracking::ErrorTrackingOptions {
+        self.options.error_tracking()
+    }
```
marandaneto, line 331: i think the panic install should be an integration

eg
options.integrations = [PanicIntegration(options, client)]

when the Client is created, it gets an option

Client(options)
client during init can go thru the `integrations` and pass `integration.install(options, self)`

so all integrations have access to the `options` and the `client`

check Java/iOS SDK
its a cleaner approach and we dont need to expose all those public APIs

src/client/async_client.rs
```diff
@@ -316,6 +316,60 @@ impl Client {
...
+
+    /// True when the calling thread is this client's transport worker thread —
+    /// the panic hook skips capturing there.
+    #[cfg(feature = "error-tracking")]
+    pub(crate) fn on_transport_worker(&self) -> bool {
+        self.transport
+            .as_ref()
+            .is_some_and(|t| t.on_worker_thread())
+    }
```
marandaneto, line 360: i dont think it should be a public API

src/client/async_client.rs
```diff
@@ -316,6 +316,60 @@ impl Client {
...
+    /// Enqueue a panic `$exception` without the tracing `capture` performs:
+    /// `capture` is `#[instrument]` and its enqueue warns once on a full queue,
+    /// both of which run subscriber code — unsafe on the already-panicking
+    /// thread. The send still happens on the worker thread.
+    #[cfg(feature = "error-tracking")]
+    pub(crate) fn enqueue_panic_event(&self, event: Event) {
+        if let Some(transport) = &self.transport {
+            transport.enqueue_panic(event);
+        }
+    }
```
marandaneto, line 371: i think we should only call `capture_panics(x)`, why do we need another public methpd?

src/error_tracking.rs
```diff
@@ -118,6 +153,155 @@ impl ErrorTrackingOptions {
...
+    if client.is_disabled() || client.on_transport_worker() {
+        return Ok(());
+    }
+    let et_options = client.error_tracking_options();
+    let event = build_panic_event(panic_info, et_options)?;
+    // Enqueue through the tracing-free path: `capture` is `#[instrument]` and
+    // warns on a full queue, both of which run subscriber code that's unsafe on
+    // the panicking thread (it could panic again -> abort, or wait on a lock the
+    // panic site holds -> hang before the previous hook runs).
+    client.enqueue_panic_event(event);
```
marandaneto, line 253: https://github.com/PostHog/posthog-rs/pull/157/changes#r3480056803
i wonder why we are being extra defensive here?

marandaneto: another alternative (which we can do across all SDKs tbh) is that our queue is a priority queue so all events that are `fatal` should be sent first (not a blocker), but something to think of

src/error_tracking.rs
```diff
@@ -529,6 +734,56 @@ fn is_internal_capture_frame(function: &str) -> bool {
         || function.contains("global::capture_exception")
 }
 
+/// Internal frames to strip from a panic stacktrace: the shared capture frames
+/// plus the panic-hook and unwinding machinery.
+fn is_internal_panic_frame(function: &str) -> bool {
```
marandaneto: this will be always outdated/brittle, i think we can do better
i dont think we should drop our own frames, we can set them as inapp = false tho

error tracking should have an in app include and exclude config, so customers can set their own thing (see eg iOS)

then we can mark normal 3rd party/rust sdk frames as in app = false as well
eg

```rust
   "std::",
   "core::",
   "alloc::",
   "backtrace::",
   "tokio::",
   "futures_core::",
   "futures_util::",
   "rust_begin_unwind",
   ...
 ```

we can also try to filter our own framas as package eg `posthog::` instead of all function names

**Comment by marandaneto** [comment]
@cat-ph left a bunch of comments/suggestions mainly to keep a cleaner public API and some 'better' behaviors imo, feel free to push back if you dont think we need that
also a few of those suggestions can be follow ups

**Comment by marandaneto** [comment]
something my agent told:

Global panic-hook tests look fragile

 The test initializes the real process-wide global client:

 ```rust
   GLOBAL_CLIENT: OnceLock<Client>
 ```

 That cannot be reset. The comment says it is the only test doing that, but this is easy to break later and can be
 order-dependent with future tests.

 Better options:

 - move global panic-hook behavior to an integration test in a separate process
 - or isolate it behind test-only reset hooks
 - or avoid asserting behavior that requires setting the real OnceLock

and:

Maybe avoid “first frame must be app frame” tests

 Tests currently assert that frame 0 is the user panic site. That encourages brittle stripping logic.

 A more robust test would assert:

 - the panic site frame exists
 - it is marked in_app = true
 - SDK/runtime frames are marked in_app = false

and:

Consider whether standalone clients should have clearer docs

 capture_panics only affects init_global. Standalone clients require:

 ```rust
   install_panic_hook(Arc<Client>)
 ```

 That is reasonable, but easy to miss. The docs/changelog should be very explicit:

 - global client: panic capture on by default
 - standalone client: no automatic install
 - manual install is process-wide and only one PostHog hook can be installed

File: src/error_tracking.rs [hunk 9]
```diff
@@ -63,6 +87,8 @@ impl Default for ErrorTrackingOptions {
             capture_stacktrace: true,
             in_app_include_paths: Vec::new(),
             in_app_exclude_paths: Vec::new(),
+            capture_panics: true,
```
ablaszkiewicz: Are we sure this should be enabled by default?

api/public-api.txt
cat-ph: tbh I thought 30s is a bit much to block the dying thread, but I'm not opposed to letting it use the default shutdown_timeout_ms if you feel that's fine too

.sampo/changesets/panic-autocapture.md
cat-ph: yeahhh that would be very useful! I looked a bit at it too, but I think we need something we can use more generic like you said, so maybe a follow-up/separate improvement 😁

api/public-api.txt
marandaneto: i think python has 10s as default
but if its too much for a panic, and i agree, makes sense

cat-ph: hmmmm good point, maybe we should "standardize" these timeouts too 😅 taking a look at other SDKs too, looks like python is actually 15s default with no flush 🤔

src/error_tracking.rs
cat-ph: fair, thanks! yeah I think we also already have `in_app_include_paths` (& exclude) + is_in_app_frame, updating this

src/error_tracking.rs
cat-ph: so this extra defensiveness was kinda intentionally dumb because capture has `#[instrument(skip(self, event), level = "debug")]` which AIUI calls into w/e the global tracing subscriber uses, 3rd party code, so if that panics at all, then aborts our panic/handling here, or worse if it does some trickery with fmt that locks the writer, it could deadlock I think 🤔

I quite like the priority queue idea!

marandaneto: mm i see
another approach is that error tracking has its own queue instead of being combined with normal events
eg logs has its own, session replay has its own, maybe error tracking should have its own as well
for them makes sense since its a different API endpoint, for errors and analycs events are the same, so not sure if it makes sense

cat-ph: yeah I think they can _keep_ the same queue but the problem here is kinda the "transport" towards the queue 😬 i.e. making sure we can't panic when enqueueing to the (events or separate) queue while we're handling an already-panic'd thread, otherwise we'll abort and lose the report; that's why the extra defensiveness here

src/client/async_client.rs
cat-ph: this is pub(crate) only, it's not in the public API

marandaneto: yep, i see `pub` and my eyes go burrr, my bad thx

api/public-api.txt
cat-ph: a bit all over the place now hmm

CleanShot 2026-06-26 at 18 15 22

marandaneto: Just go with what makes sense here and i deal with this after

src/error_tracking.rs
cat-ph: not _entirely_ sure, I think @marandaneto suggested it in the previous impl of the PR (before the transport change), and it also makes sense IMO, although most other SDKs don't do that 🤔 🤔 🤔 (https://github.com/PostHog/posthog-rs/pull/157#discussion_r3482506134)

I don't think it's _very_ common to get panics in rust compared to maybe js uncaught exceptions or similar so I wouldn't expect to add volume, the only concern is maybe the panic hook itself being auto-registered

src/client/async_client.rs
cat-ph: it's also `pub(crate)` not directly public, I think it's a bit "cleaner" but can also drop it if you want

.sampo/changesets/panic-autocapture.md
cat-ph: ah I saw you actually opened https://github.com/PostHog/posthog-rs/issues/151 - I think that encompasses this pretty much, added a comment to refer to this too

cat-ph:
I resolved some previous thread pre-emptively, but some updates:

- changed the level to fatal
- updated the in_app classification / frame stripping logic
- fixed the test fragility

and them some remaining discussions/thoughts:

- extra defensiveness in the enqueue path - _kinda_ intended to try and keep code that could panic to the minimum
- whether we should have a separate, lower `panic_flush_timeout_ms` - I'd say yes, but if we want to align timeouts across SDKs and think this may not be needed, maybe we can 'hardcode' this to 2s for now and not have it public (we can always just expose it later on if we need to), open to thoughts
- whether enabled by default, IMO yes?

src/client/async_client.rs
cat-ph: I like the approach, I looked at it a bit, was a bit trickier with the lifetimes (since we allow the global client + standalone clients, maybe we should just keep the global one to be simpler 🤔), do you think this would be better as a follow-up?

src/client/async_client.rs
marandaneto: all good i just misunderstood the `pub(crate)`

src/client/async_client.rs
marandaneto: yep makes sense, tbh i think only the global is a better approach
i dont even know if its even possible to set up multiple hooks
so by default only the global instance and allowing to choose the instance can be def a follow up
this was just a suggestion tho, felt simpler to me, your call here if we go this way

src/error_tracking.rs
marandaneto: no string opinions tho, i think can be disabled by default as well, if thats what all the other sdks do

marandaneto:
> I resolved some previous thread pre-emptively, but some updates:
> 
> * changed the level to fatal
> * updated the in_app classification / frame stripping logic
> * fixed the test fragility
> 
> and them some remaining discussions/thoughts:
> 
> * extra defensiveness in the enqueue path - _kinda_ intended to try and keep code that could panic to the minimum
> * whether we should have a separate, lower `panic_flush_timeout_ms` - I'd say yes, but if we want to align timeouts across SDKs and think this may not be needed, maybe we can 'hardcode' this to 2s for now and not have it public (we can always just expose it later on if we need to), open to thoughts
> * whether enabled by default, IMO yes?

cool stuff, lets keep that way then

src/error_tracking.rs
```diff
@@ -637,18 +915,30 @@ fn default_in_app_function(function: &str) -> bool {
             .split("::")
             .next()
             .unwrap_or_default(),
-        "alloc" | "backtrace" | "core" | "posthog_rs" | "reqwest" | "std" | "tokio"
+        "alloc"
```
marandaneto, line 896: worth adding:

- tracing_core::
 - tracing::
 - log::
 - anyhow::

marandaneto:
Add a test that the actual panic-site frame is in_app = true

 The current test checks:

 - panic site frame exists
 - panic/unwind machinery is in_app = false

 But it does not verify that the user panic-site frame is marked in_app = true.

 I’d ask for this, ideally in tests/test_panic_global.rs, because that panic originates from an integration-test crate,
 not inside posthog_rs itself.

 Something like:

 ```rust
   frames.iter().any(|frame| {
       frame["function"].as_str().is_some_and(|f| f.contains("...panic site..."))
           && frame["in_app"] == true
   })
 ```

src/error_tracking.rs
```diff
@@ -55,6 +62,23 @@ pub struct ErrorTrackingOptions {
...
+    /// yourself; the flag does not affect standalone clients.
+    capture_panics: bool,
+    /// How long the installed panic hook blocks the panicking thread waiting for
+    /// the `$exception` to be sent before letting the panic proceed (default:
+    /// 2000 ms). Deliberately short and separate from `shutdown_timeout_ms`: the
+    /// hook runs on the dying thread, so a long wait would freeze the crash —
+    /// and delay the panic message, which only prints after the flush — whenever
+    /// PostHog is slow or unreachable. Only consulted when the panic hook is
+    /// installed via [`install_panic_hook`].
+    panic_flush_timeout_ms: u64,
```
marandaneto: panic_flush_timeout_ms doc is still wrong

 Comment still says:

 │ Only consulted when the panic hook is installed via install_panic_hook.

 But the global hook installed by init_global also uses it.

 Suggested wording:

 │ Consulted by the panic hook, whether installed automatically via init_global or manually via install_panic_hook.

marandaneto (approved):

marandaneto:
@cat-ph great work and nice addition
left a few comments/non blockers (suggestions), but happy as is!

src/error_tracking.rs
cat-ph: I left it out for now from the public API, keeping just the global one to simplify!

src/error_tracking.rs
cat-ph: actually yeah, let's go with disabled by default since all others do that too, we can always enable later!

src/client/async_client.rs
cat-ph: let's keep global only for now to be simpler, I think it makes more sense, we can add per-client separate and we can follow-up with the integration approach then 😁

cat-ph:
aight last few small changes to simplify actually, I think it's better:

- disable by default (what other SDKs do)
- hardcode panic flush timeout (can make it configurable later or align with other SDK timeouts)
- keep global only, simplifies the public API, we can add per-client later if others find it useful

will test again thoroughly tomorrow (today? 😆) & merge unless any objections 🚀

src/error_tracking.rs
```diff
@@ -118,6 +155,131 @@ impl ErrorTrackingOptions {
...
+#[allow(deprecated)]
+fn build_panic_event(
+    panic_info: &panic::PanicInfo<'_>,
+    et_options: &ErrorTrackingOptions,
+) -> Result<Event, Error> {
+    let exception = Exception::from_panic_info(panic_info, et_options.capture_stacktrace());
+
+    let mut event = Event::new_anon("$exception");
+    if let Some(location) = panic_info.location() {
+        event.insert_prop("$exception_panic_file", location.file())?;
```
hpouillot, line 275: Just to understand, why adding those properties ? They won't be displayed on the UI and they are not processed, are they ?

PR #158, opened 2026-06-25
feat: capture native instruction addresses and debug images

cat-ph:

PostHog error tracking symbolicates native (Rust/C/C++) stack frames server-side. For that the SDK attaches, per captured exception, each frame's absolute **instruction address** and an event-level **`$debug_images`** list (the loaded modules), so the backend (cymbal) can map addresses → source against debug symbols uploaded with `posthog-cli`. This enriches **both** manual `capture_exception` and automatic panic captures.

What's captured:
- Per-frame `instruction_addr` (plus `symbol_addr`/`image_addr` when known); frames tagged `platform: "native"`.
- `$debug_images`: `type`, `debug_id`, `code_id`, `image_addr`, `image_size`, `image_vmaddr`, `code_file`, normalized `arch`.
- Debug ids match the server/CLI (`symbolic`): GNU build id on ELF, `LC_UUID` (dSYM-cased) on Mach-O, GUID+age on Windows PDB.
- Addresses and debug-image linkage are omitted when no uploadable debug image matches a frame, so events stay clean for stripped builds.
- A resolvable frame emits one entry per physical address and lets cymbal expand its inline chain; only non-resolvable frames keep client-side inline expansion, so inlines aren't double-expanded server-side.

Gated behind the default-on `error-tracking` feature; adds `findshlibs` for loaded-module enumeration.

## some screenshots of how those look like

<details>
<summary>no debug symbols, no upload</summary>

regular
CleanShot 2026-07-01 at 01 45 04

panic
CleanShot 2026-07-01 at 01 50 52

</details>

<details>
<summary>(no & yes) debug symbols, uploaded symbol sets => source context</summary>

fwiw debug symbols / no debug symbols look the same (symbolicate + source context from upload)

regular
CleanShot 2026-07-01 at 01 53 02

panic
CleanShot 2026-07-01 at 01 56 33

</details>

<details>
<summary>debug symbols, no upload => symbolicated, no source context</summary>

regular
CleanShot 2026-07-01 at 01 58 13

panic
CleanShot 2026-07-01 at 02 03 20

</details>

Agent-authored, so listing only what was actually run + e2e test on a devbox:
- `cargo test` green for `error-tracking` and `error-tracking,async-client` (unit + integration; the live e2e test stays `#[ignore]`).
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean across both combos.
- Native unit tests cover GNU-build-id → debug-id conversion (endianness-gated), address-range module matching, debug-image referencing, and resolvable-vs-stripped frame shape.
- `api/public-api.txt` matches `main` (native types are `pub(crate)`).

## :pencil: Checklist

- [x] Ran `sampo add` to generate a changeset file (`posthog-rs: minor`).

## :robot: Agent notes

Agent-authored. Debug-id computation matches `symbolic`'s `ElfObject::compute_debug_id`, including gating the GNU-build-id field swap on target endianness so big-endian ELF targets are correct.


src/error_tracking.rs
```diff
@@ -655,17 +904,72 @@ fn capture_frames_current_first(skip: usize) -> Vec<StackFrame> {
                 return;
             }
 
+            layers.push((filename, symbol.lineno(), function.unwrap_or_default()));
+        });
+
+        if resolvable {
+            // The server can symbolicate this address, so emit ONE frame per
+            // physical frame carrying the raw `instruction_addr` and let the
```
cat-ph, line 913: @hpouillot curious to hear your thoughts for this approach, it's not going to be as bad here in rust, but will matter even more in go IMO..

so we do one frame per physical frame if it's resolvable, hoping cymbal finds a symbol set and can symbolicate, and cymbal expands the inlined frames. if we also expanded it here, cymbal would duplicate

but if there's no symbol set uploaded, we 'lose precision' because we could've expanded the inlined frames already client-side..

given go will be even more client heavy instead of cymbal symbolication, not sure if we wanna keep this same approach or should we either try to deduplicate in cymbal somehow after symbolication _OR_ even better (but I gotta play with this to make sure it works), during symbolication group frames by address, discard potentially

[this is still technically likely a follow-up, current PR works as-is but thinking ahead fwiw]

src/error_tracking.rs
```diff
@@ -613,25 +648,230 @@ impl ExceptionStacktrace {
...
+    end: u64,
+    is_main: bool,
+    image: DebugImage,
+}
+
+const fn native_image_type() -> &'static str {
+    if cfg!(any(target_os = "macos", target_os = "ios")) {
+        "macho"
+    } else if cfg!(target_os = "windows") {
+        "pe"
```
cat-ph, line 721: caveat is, I've not _properly_ tested windows yet, just elf and mach-o 😅

Cargo.toml
```diff
@@ -22,6 +22,7 @@ semver = "1.0.24"
 derive_builder = "0.20.2"
 uuid = { version = "1.13.2", features = ["serde", "v7"] }
 backtrace = { version = "0.3.74", optional = true }
+findshlibs = { version = "0.10", optional = true }
```
marandaneto, line 25: is that also under the error tracking feat? so just analytics/flags dont get this dep for no reason

marandaneto: ah just saw L55 my bad

src/error_tracking.rs
marandaneto: shouldnt we just do all the work server side because it has less of a performance impact on the hosting app, and it's easier to fix things if its broken

cat-ph: tbh I think the end goal is:

- client-side we send richest info we have/can
- server-side (cymbal) we dedup as needed and enrich

a lot of people use rust with some debug symbols (for tracing/profiling too etc) so doing & sending it here means they get useful frames out-of-the-box (even before they upload symbol sets, but that will help with source context so they should probably still do it)

imo it's even more a golang issue since there it'll be pretty much all client-side resolved, so we _have_ to do something like this to not lose info

marandaneto: i see, so its possible to do stuff at runtime if symbols arent stripped
so users get value ootb if not uploading symbols
agree

the problem is if we do that
they upload symbols
stuff gets overwritten server side?

we need to decide which one wins, like, who symbolicates better, is that right?

cat-ph: yes! that's the tricky part, ideally they symbolicate the same, BUT there's an edge case here:

when inlining functions, there's a 'matrix' here 

| client-side | server-side | result |
|--|--|--|
| one frame per physical frame | no symbols uploaded | losing the logical frames because we downgraded client |
| one frame per logical frame (expanded inlining) | no symbols uploaded | no precision loss |
| one frame per physical frame | symbols uploaded | no precision loss, cymbal expands physical frame |
| one frame per logical frame | symbols uploaded | cymbal would *duplicate* the physical->logical frames by expanding, so extra duplicate frames |

ideally we send one per logical frame client-side, and deal with it in cymbal without the duplicate, so we always have the most data, but this is the reason we 'downgrade' here to physical frames

it's also how it works for the ios right now already AIUI

(I'm trying now separately to do my grouping frames per address idea in cymbal to prevent this)

marandaneto: maybe you just need an extra metadata eg `$frame.resolved = true|false` client side if there was builtin symbols and then cymbal just try to do its work if `frame.resolved = false`? then theres no duplicates at all

cat-ph: I thought about a `client_resolved` or similar (since we already have a `resolved` one that's mostly just output post-cymbal, but naming tbd), it also gets tricky with source context, cause we'd want to enrich source context even if we could symbolicate client-side (_but_ that's still doable, maybe simpler 🤔 might spare cymbal of the work if it knows it's already resolved)

marandaneto: makes sense

src/error_tracking.rs
```diff
@@ -655,17 +904,72 @@ fn capture_frames_current_first(skip: usize) -> Vec<StackFrame> {
...
                 synthetic: false,
-                resolved: true,
-                platform: "rust".to_string(),
+                platform: "native".to_string(),
+                instruction_addr: Some(format!("0x{instruction_addr:x}")),
+                symbol_addr: (frame_symbol_addr != 0).then(|| format!("0x{frame_symbol_addr:x}")),
+                image_addr: module.map(|m| m.image.image_addr.clone()),
             });
-        });
+        } else if layers.is_empty() {
```
hpouillot: I am wondering if we want to keep this ? the frame looks mostly empty I am not sure it will bring a lot of value and / or might confuse fingerprint if the function name is the same

src/error_tracking.rs
hpouillot: If we can make the product useful without symbolset upload, I think we should do it but at the same time if we can reduce data transfer, we should also do it. Right now, I don't think there is a way to know client-side if symbol sets were uploaded, so it means we will need to always symbolicate client-side when we can ? Maybe we can add an option to disable symbolication client side ? (for later)

In any case it makes to add a bool at the frame level 👍

src/error_tracking.rs
```diff
@@ -655,17 +904,72 @@ fn capture_frames_current_first(skip: usize) -> Vec<StackFrame> {
...
+            // so the server can't symbolicate this address. Keep the client-side
+            // inline expansion (one frame per layer, no native addresses) — it's
+            // the only way these inlined calls survive.
+            for (filename, line_no, function) in layers {
+                frames.push(StackFrame {
+                    filename,
+                    line_no,
+                    function,
+                    lang: "rust".to_string(),
+                    in_app: false,
```
hpouillot, line 953: Why not using the `module.is_some_and(|m| m.is_main)` here ?

hpouillot: Maybe we can even use this function, as most of the lib will be bundled in the main module ?https://github.com/PostHog/posthog-rs/blob/960aa576919f1fbc00bcab8ddb653732b0121d5f/src/error_tracking.rs#L876

src/error_tracking.rs
cat-ph: hmm yeah, I think we can drop it tbh, good point

src/error_tracking.rs
cat-ph: this field here is not _really_ accurate because it gets overwritten by `write_into` above (around lines 567) because it has names, the same as the in_app false in the resolvable case. since I'm dropping the empty frame above, I think it's safe to keep both as in_app false, I'll add a comment that they get replaced

(the `layers.is_empty()` was the odd one out using `module.is_some_and(|m| m.is_main)` because it didn't have names 😁)

src/error_tracking.rs
cat-ph: > Right now, I don't think there is a way to know client-side if symbol sets were uploaded, so it means we will need to always symbolicate client-side when we can

_exactly_!

I like the idea of allowing people to disable client-side symbolication 🤔 but I think we still need to gather these to be able to enrich source context on cymbal so I'm not sure if it'll save a lot tbh 😓 

edit: _maybe_ actually filename should be enough for source context

cat-ph: added a `client_resolved` flag, will work on cymbal separately to use it

hpouillot (approved):

marandaneto (approved):

PR #159, opened 2026-06-25
feat: add on_error client hook for terminal batch delivery failures

eli-r-ph:

This PR adds an additive, observability-only `on_error` hook so callers can react to terminal delivery failures without reverting to a blocking `capture()`. Purely additive change. It fires only if a handler is registered by the caller, and only when a response is non-2xx or all retries are exhausted. The v1 variant will expose the rich detail that endpoint returns so the caller can debug event delivery issues quickly and at the granularity they want.

The background transport (introduced in 0.14.0) made `capture()`/`capture_batch()` fire-and-forget. That decoupled the API call from network I/O, but it also removed the only signal callers had that a batch was *permanently* lost — a permanent reject (e.g. 400/402) or a transient failure that exhausts the retry budget. Today the SDK only emits a `WARN` and drops the events; there is no programmatic hook to log, count, or alert on dropped data.

- New `ClientOptionsBuilder::on_error(...)`, plus public `OnErrorHook` and `CaptureFailure` types.
- `CaptureFailure` exposes `error()` (the batch cause), `event_count()`, `attempt()`, and `historical_migration()`.
- The hook fires **once per terminal batch** on the background worker:
  - permanent reject, or exhausted retry budget (`error()` is `Some`);
  - on the V1 capture pipeline, also when a request was `2xx` but per-event verdicts (`retry`/`drop`) leave events unpersisted after the retry budget (`error()` is `None`). In that case the feature-gated `CaptureFailure::event_results()` exposes the per-event server verdicts.
- Registering at least one hook silences **only** the default drop `WARN` it replaces; all other drop warnings (shutdown, serialization, queue-full) are unchanged.
- Hook panics are caught so a misbehaving hook cannot wedge the worker thread; multiple hooks fire in registration order.

This is purely additive — no existing signature changes. V1-specific behavior is implemented and tested but its public rustdoc is `cfg`-gated (invisible on the default docs.rs build); full v1 documentation is deferred to the v1 cutover.

New unit tests in `src/client/transport.rs` (driven by the existing `ManualClock`/`httpmock` harness) and `src/client/common.rs`, plus a public-API integration test in `tests/test_transport.rs`. Coverage:

- fires exactly once on a permanent reject and on retry exhaustion (correct `attempt`/`error`/`event_count`);
- does **not** fire on success or on a retry that eventually succeeds;
- V1: fires on a `drop` and on an exhausted `retry` *within* a 2xx (`error()` is `None`, `event_results()` populated), and does **not** fire for `ok`/`warning` verdicts;
- hooks run in registration order and a panicking hook is isolated; the worker survives and keeps delivering.

Ran the full CI matrix locally (default, `capture-v1`, the `--no-default-features` combos, and the `e2e-test` config) — all green — plus `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo doc`, and regenerated `api/public-api.txt` via `scripts/check-public-api.sh --update`.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -1,11 +1,18 @@
 use std::sync::{Arc, Mutex};
 
 use crate::endpoints::{EndpointManager, DEFAULT_HOST};
+use crate::error::Error;
 #[cfg(feature = "error-tracking")]
 use crate::error_tracking::ErrorTrackingOptions;
 use crate::event::Event;
+#[cfg(feature = "capture-v1")]
+use crate::event_v1::EventResult;
 use derive_builder::Builder;
+#[cfg(feature = "capture-v1")]
+use std::collections::HashMap;
 use tracing::warn;
+#[cfg(feature = "capture-v1")]
+use uuid::Uuid;
 
 mod common;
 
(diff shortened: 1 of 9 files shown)


eli-r-ph:
Thanks @greptile-apps — addressed both in 09a702d:

1. **`event_results()` includes `ok`/`warning`**: intentional — the map is meant to expose the full per-event server response so callers can introspect every verdict, not just losses. Rather than filter it down (which would discard that signal), I clarified the docs: `event_results()` now documents that it reports all verdicts and can be larger than `event_count()` (which counts only `retry`/`drop`), and `event_count()` cross-references it. Added a V1 mixed-verdict test (`on_error_v1_event_results_includes_persisted_verdicts`: ok + warning + drop) asserting `event_count == 1` while `event_results()` carries all three.

2. **Parameterize the transport tests**: collapsed the two "fires once on terminal failure" tests into one table-driven `on_error_fires_once_on_terminal_failure` (permanent-reject vs exhausted-retries). Left the two "not fired" tests separate — they have genuinely different orchestration (one swaps the mock mid-flight to a 200 to prove a retry that succeeds does not fire), so a shared table would obscure rather than clarify.

Note: the failing `Feature_Flags` compliance checks are unrelated to this PR — they are `404`s from the test adapter `/get_feature_flag` route; the v0/v1 capture suites are 29/29 and 94/94 green.

src/client/mod.rs
```diff
@@ -87,6 +94,107 @@ impl BeforeSendHook {
...
+pub struct OnErrorHook(SharedOnErrorHook);
+
+impl OnErrorHook {
+    /// Create a new on-error hook.
+    pub fn new<F>(hook: F) -> Self
+    where
+        F: FnMut(&CaptureFailure<'_>) + Send + 'static,
+    {
+        Self(Arc::new(Mutex::new(Box::new(hook))))
+    }
```
dustinbyrne, line 117: it doesn't look like callers need to reference or instantiate `OnErrorHook`. can the struct/`new` visibility be downgraded to `pub(crate)`?

src/client/transport.rs
```diff
@@ -725,10 +729,50 @@ impl Pipeline {
...
+                                results: &batch.final_results,
+                            },
+                        );
+                    }
+                }
+            }
             Step::Fail(e) => {
-                warn!("posthog-rs: dropping {} event(s): {e}", batch.pending.len());
-                dec_len(&self.len, batch.pending.len());
+                let count = batch.pending.len();
```
dustinbyrne, line 759: minor: a mixed retry + drop response followed by a `Step::Fail` result can make `event_count()` under-report total lost events, because earlier drop verdicts are removed from pending and only remain in `event_results()`.

dustinbyrne (commented):
couple of small comments below, otherwise looks good

marandaneto (changes requested):
i think we should make this more generic since we have more APIs and more types of errors
@ioannisj since hes planning to work on SDK observability
otherwise we'd need one hook per API call, or error type
JS has `on_request_error`
RN has:
 ```ts
   const unsubscribe = posthog.on('error', (err) => {
     console.log('PostHog SDK error', err)
   })
 ```

so not sold on either of those but i think we'd need to think about that first
whats your goal here @eli-r-ph ?
what will you do with this data?
we should also not emit the errors and then capture the error ourselves, this would go into an endless loop if our SDK is buggy or having an incident

ioannisj:
> i think we should make this more generic since we have more APIs and more types of errors @ioannisj since hes planning to work on SDK observability otherwise we'd need one hook per API call, or error type JS has `on_request_error` RN has:
> 
> ```ts
>   const unsubscribe = posthog.on('error', (err) => {
>     console.log('PostHog SDK error', err)
>   })
> ```
> 
> so not sold on either of those but i think we'd need to think about that first whats your goal here @eli-r-ph ? what will you do with this data? 

Agreed on making this hook more generic. I think it's already wide enough (we can call on_error from other places as well), we maybe just need to rename `CaptureFailure` to a more neutral type?

> we should also not emit the errors and then capture the error ourselves, this would go into an endless loop if our SDK is buggy or having an incident

I don't see how this PR will suffer from this since this is just calling a non-blocking hook for SDK use observability. For now I think we are okay here

eli-r-ph:
👋 hey folks. I am just hoping to restore the error visibility we had prior to the transport change in the 0.14.x release. My e2e smoke tests broke for capture v1 on this release because of the loss of client-side error exposure. 

I think the error hook pattern is nicer than what we had prior to the transport change anyway, well suited to the new queued async batch submissions in this SDK, and would allow us to expose more actionable/granular fail modes to the end user that capture v1 offers.

A more generic error handler for the whole SDK makes sense to me too. If we could at least surface a request-level error, it gets us back to parity with what we had before on 0.13.x release, and would solve my immediate testing gap.

I agree with @ioannisj about the feedback loop risk. `posthog-go` already has a similar mechanism with `Callback.Failure` and no one's run afoul of the loop risk yet AFAIK (?) but I agree it's worth documenting these risks

marandaneto:
> I don't see how this PR will suffer from this since this is just calling a non-blocking hook for SDK use observability. For now I think we are okay here

sdk has a transport bug, or really any bug that causes this to be triggered N times

```
posthog.init (..., onError { error ->

  posthog.capture(error)

})
````

this will be an endless loop

eli-r-ph:
👋 Looks like we have a good idea how to proceed to make this more generally useful now. I'm going to close this out and open a new PR with the proposed changes for @ioannisj and @marandaneto to have a look at

eli-r-ph:
Superseded by #167, which generalizes the `on_error` hook to cover all SDK network surfaces (capture, remote /flags, and the local-evaluation poller) via a single `PostHogError` enum, rather than capture-only.

PR #161, opened 2026-06-27
fix: respect Retry-After minimum delay

marandaneto:

`Retry-After` is a server-provided minimum retry delay. The Rust SDK previously used numeric `Retry-After` values as a direct override, which could retry earlier than a custom configured backoff, and it ignored valid HTTP-date values.

This aligns retry behavior with HTTP semantics and the Go SDK by waiting for the longer of the configured backoff and `Retry-After`, and by accepting both delay-seconds and HTTP-date forms without adding a new dependency.

- `cargo fmt --check`
- `cargo test --lib`
- `cargo test --features capture-v1 --lib`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/src/client/retry.rs
+++ b/src/client/retry.rs
@@ -1,7 +1,8 @@
 //! Runtime-agnostic retry primitives shared by the V0 and V1 capture paths.
 
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 
+use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
 use reqwest::header::HeaderMap;
 
 use super::ClientOptions;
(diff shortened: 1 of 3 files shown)


marandaneto:
@eli-r-ph found this dif. between rust and go, pls check

Cargo.toml
```diff
@@ -33,6 +33,7 @@ tokio = { version = "1", features = [
     "macros",
 ], optional = true }
 flate2 = "1.0"
+httpdate = "1.0"
```
marandaneto: can we implement our own parsing to avoid an extra dependency? if its simple enough

marandaneto: Good call! Fixed in f244f8e — removed the direct `httpdate` dependency and parse HTTP-date forms using the existing `chrono` dependency plus small local fallback parsing.

src/client/retry.rs
```diff
@@ -182,14 +204,23 @@ mod tests {
...
     }
 
+    #[test]
+    fn backoff_configured_delay_wins_when_retry_after_is_shorter() {
+        let opts = test_opts();
+        assert_eq!(
+            backoff_duration(&opts, 1, Some(Duration::from_millis(1))),
+            Duration::from_millis(100)
+        );
+    }
```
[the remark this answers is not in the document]
marandaneto: Thanks, applied in f244f8e — folded the Retry-After min-delay cases into one table-driven test and added an exact-tie case.

src/client/retry.rs
```diff
@@ -1,7 +1,8 @@
 //! Runtime-agnostic retry primitives shared by the V0 and V1 capture paths.
 
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
 
+use chrono::{DateTime, NaiveDateTime, TimeZone, Utc};
```
eli-r-ph, line 5: 👍 using `Duration` here is nicer

eli-r-ph (approved):
Good change, reframing `Retry-After` as a min not a cap LGTM 🙇 thanks! We are also now returning date headers in responses from capture (and Envoy on gateway 503s)

marandaneto:
> Good change, reframing `Retry-After` as a min not a cap LGTM 🙇 thanks! We are also now returning date headers in responses from capture (and Envoy on gateway 503s)

mm do SDKs read those? can be a follow up if not yet

PR #162, opened 2026-06-27
fix: route flag called events through capture transport

marandaneto:

`$feature_flag_called` events were special-cased in Rust for V1 capture: they bypassed the normal transport, sent immediately, and did not retry. This made their routing/retry semantics differ from normal capture events.

This PR routes feature flag called events through the same capture transport as all other events so they are batched, flushed, and retried consistently.

- `cargo test`
- `cargo test --features capture-v1`
- `cargo test --no-default-features --features capture-v1 --test test_evaluate_flags`
- `cargo test --no-default-features --test test_evaluate_flags`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.sampo/changesets/doughty-thunderbearer-otso.md
+++ b/.sampo/changesets/doughty-thunderbearer-otso.md
@@ -1,0 +1,5 @@
+---
+cargo/posthog-rs: patch
+---
+
+Route feature flag called events through the normal capture transport.
--- a/src/client/async_client.rs
+++ b/src/client/async_client.rs
@@ -4,8 +4,7 @@ use std::error::Error as StdError;
 use std::sync::{Arc, OnceLock};
 use std::time::Duration;
 
-use reqwest::header::USER_AGENT;
-use reqwest::{header::CONTENT_TYPE, Client as HttpClient};
+use reqwest::{header::CONTENT_TYPE, header::USER_AGENT, Client as HttpClient};
 use serde_json::json;
 use tracing::{debug, instrument, trace, warn};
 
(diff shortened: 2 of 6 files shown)


marandaneto:
@eli-r-ph found this dif. between rust and go, pls check

eli-r-ph (approved):
Change LGTM - I deferred messing with this for the next stage of v1 rollout (removing the old plumbing) because I wanted to think about whether we should combine this very high-volume, low-value, fire-and-forget event type into the main capture v1 submission path where higher-value events are submitted, and more aggressive retries will be the norm etc.

Due to the high volume of `$feature_flag_called` events submitted at the moment from all SDKs, we just added a deduplication layer just for these to the ingestion pipeline backend last week. On the teams it's enabled for, it's dropping >80% of them.

That said, the risk may be overstated here, and the change itself LGTM - your call whether we should land this and try it out 👍

marandaneto:
> Change LGTM - I deferred messing with this for the next stage of v1 rollout (removing the old plumbing) because I wanted to think about whether we should combine this very high-volume, low-value, fire-and-forget event type into the main capture v1 submission path where higher-value events are submitted, and more aggressive retries will be the norm etc.
> 
> Due to the high volume of `$feature_flag_called` events submitted at the moment from all SDKs, we just added a deduplication layer just for these to the ingestion pipeline backend last week. On the teams it's enabled for, it's dropping >80% of them.
> 
> That said, the risk may be overstated here, and the change itself LGTM - your call whether we should land this and try it out 👍

another approach we could think of would be a priority queue for events
this makes the sdk much simpler tho, so happy to get this in and find another alternative if the dedupe server side does not go well which it seems like its going well

PR #167, opened 2026-06-30
feat: general on_error observability hook

eli-r-ph:

Supersedes #159. That PR added an `on_error` hook scoped to terminal **capture** batch failures on the transport worker. This generalizes the hook to cover every network surface the SDK talks to, so callers get one place to observe terminal failures without giving up the non-blocking API:

- **Capture** — a batch the SDK gave up delivering (permanent reject, exhausted retries, serialization failure), or on the `capture-v1` pipeline a `2xx` whose per-event verdicts left events unpersisted after the retry budget.
- **Feature flags** — a failed remote `/flags` request (transport error, non-success status, or unparseable body). Fires at the leaf request, so it still reports when `evaluate_flags` degrades gracefully to local results.
- **Local evaluation** — a failed definitions poll (initial load and background refresh).

Design choices:

- One `#[non_exhaustive]` `PostHogError<'a>` enum with a variant per surface, each carrying a `#[non_exhaustive]` failure struct read through accessors. New surfaces and fields can be added without breaking callers.
- The failure is passed by reference and borrows from the caller, so the common hookless path stays allocation-free; response bodies are only retained when a hook is registered.
- The hook is **observability-only**. Docs and the example warn loudly: never call back into the SDK from it (`capture*`/`flush`/`shutdown`) — emitting while handling a capture failure forms an amplification loop, and re-entering can deadlock the hook's mutex. Panics are caught and ignored.
- `LocalEvaluationConfig` is public and not `#[non_exhaustive]`, so hooks are stored on the (private) pollers and injected by the client builder rather than added to the config struct — no breaking change.

Drive-by fix: the blocking local-eval poller's background thread slept the full poll interval before checking its stop flag, so `stop`/`Drop` could block for up to a whole interval. It now wakes in short steps.

New public API (additive): `PostHogError`, `CaptureFailure`, `FlagsFailure`, `LocalEvaluationFailure`, `ClientOptionsBuilder::on_error`, and (under `capture-v1`) `V1ErrorResponse`. `api/public-api.txt` regenerated.

- New `tests/test_on_error.rs` integration coverage (blocking + async) for the flags and poller surfaces: status/endpoint/distinct-id/body propagation, firing on the degrade-to-local path, and initial poll failure.
- Capture-surface firing covered by transport unit tests.
- Full CI feature matrix (`default`, `capture-v1`, `no-default-features` combos) green; `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo doc -D warnings` clean.
- Compliance adapter builds for v0 and v1; capture behavior is unchanged when no hook is registered.

## :pencil: Checklist

### If releasing new changes

## :robot: Agent context

Generated with assistance from Cursor. The agent designed the unified `PostHogError` enum, wired firing sites across `transport.rs` / `async_client.rs` / `blocking.rs` / `local_evaluation.rs`, added the integration tests, regenerated the public-API snapshot, and authored the changeset. Reviewer should sanity-check the ownership/borrowing on the capture firing path and the observability-only guarantees.


src/client/transport.rs
```diff
@@ -764,11 +766,14 @@ impl Pipeline {
         let payload = match serde_json::to_vec(&req) {
             Ok(p) => p,
             Err(e) => {
-                warn!(
-                    "posthog-rs: dropping {} event(s), serialization failed: {e}",
-                    batch.pending.len()
-                );
-                dec_len(&self.len, batch.pending.len());
+                let count = batch.pending.len();
+                warn!("posthog-rs: dropping {count} event(s), serialization failed: {e}");
```
ioannisj: Should we gate this on `on_error.is_empty()` as well per doc [here](https://github.com/PostHog/posthog-rs/pull/167/changes#diff-1b84118fef4e05f6f4c73c9ad3c5d469ec065f0ef62e20448683dadf6f281c6dR371)?

src/client/on_error.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub(crate) status: Option<u16>,
+    pub(crate) attempt: u32,
+    pub(crate) event_count: usize,
+    pub(crate) historical_migration: bool,
+    #[cfg(feature = "capture-v1")]
+    pub(crate) request_id: Option<&'a Uuid>,
+    #[cfg(feature = "capture-v1")]
+    pub(crate) results: &'a HashMap<Uuid, EventResult>,
+    #[cfg(feature = "capture-v1")]
+    pub(crate) error_response: Option<&'a V1ErrorResponse>,
```
ioannisj, line 103: Since this is now public, should it be also decorated with `#[non_exhaustive]` similar to other public types?

src/client/on_error.rs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        Self(Arc::new(Mutex::new(Box::new(hook))))
+    }
+
+    /// Invoke the hook. Recovers a poisoned mutex (a prior hook panic) so one
+    /// bad invocation doesn't permanently wedge the hook.
+    pub(crate) fn apply(&self, failure: &PostHogError<'_>) {
+        let mut hook = self
+            .0
+            .lock()
```
ioannisj: Won't this deadlock if users re-enter the SDK call path on the same thread within the hook (even though we mention in changeset that they should not)?

ioannisj (approved):
LG to me even though can't follow rust code 100%. Left a couple of comments but would also want @marandaneto eyes as well?

marandaneto:
looks good, nice!
We'll soon have more endpoints, such as /logs and /ai maybe at some point remote config support. Is the current implementation extensible so we can keep adding new endpoints while using the same hook signature, since it's not exhaustive?
if thats the case, besides a few comments @ioannisj already made, lgtm

src/client/on_error.rs
eli-r-ph: I'll take a look at options here 👍 good callout. the looping hook triggers is pathological either way, will be a risk we carry either way

src/client/on_error.rs
eli-r-ph: good call, will do 👍

src/client/transport.rs
eli-r-ph: checking

src/client/on_error.rs
eli-r-ph: updated, removed the locking/deadlock issue. Leaning on docs updates for warning on pathological cases 👍

src/client/on_error.rs
eli-r-ph: done!

src/client/transport.rs
eli-r-ph: done, thanks!

PR #168, opened 2026-07-01
Exclude eyre from default in-app frames & add eyre example

ranile:

`eyre` is a widely used error handling like `anyhow`, that many people prefer to use. 

This PR:
1. Adds `eyre` example to `examples/error_tracking.rs`. This example is relevant to `anyhow` as well 
2. Extends `default_in_app_function` so `eyre`, `color_eyre`, and `stable_eyre` (all under same umbrella). `eyre` frames are excluded from in-app classification, similar to how `anyhow` is already handled. 

PS: the anyhow example at https://posthog.com/docs/error-tracking/installation/rust is broken, which prompted me to dig deeper and write this PR

`cargo check` & `cargo test`. I did not run the example with the real posthog key and account

## :pencil: Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -809,6 +809,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "eyre"
+version = "0.6.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7cd915d99f24784cdc19fd37ef22b97e3ff0ae756c7e492e9fbfe897d61e2aec"
+dependencies = [
+ "indenter",
+ "once_cell",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ranile:
> `stable_eyre` is missing a representative frame in the test. `eyre` has `"eyre::Report::msg"` and `color_eyre` has `"color_eyre::config::EyreHook::into_eyre_hook::{{closure}}"`, but `stable_eyre` has no entry. If the crate name is ever spelled differently at the Rust symbol level (e.g. hyphens vs underscores), the omission means the exclusion is untested. Adding something like `"stable_eyre::handler::DefaultHook"` (or any real `stable_eyre` frame) would complete the coverage and match the pattern already established for the other two crates.

stable_eyre only touches how backtrace is handled. It does not touch handling/propagation of errors. That's still delegated to `eyre`'s types: https://github.com/eyre-rs/stable-eyre/blob/48f50e37665195a77e8e2dc53fd0bb1201a9a332/src/lib.rs#L97-L133

hpouillot (approved):
lgtm

marandaneto:
@hpouillot without a changeset this will only be available in the next version

PR #174, opened 2026-07-02
feat: add secret_key config, deprecate personal_api_key

turnipdabeets:
## Problem

The credential used for local feature flag evaluation and remote config accepts **either** a Personal API Key (`phx_...`) **or** a Project Secret API Key (`phs_...`). The builder field/method was named `personal_api_key`, which is misleading since a Project Secret API Key is equally valid (and preferred).

## Change

- Add a canonical `secret_key(...)` builder method / config field. It accepts either a Personal API Key or a Project Secret API Key, and is the value used for local flag evaluation and remote config.
- Keep `personal_api_key(...)` as a `#[deprecated]` alias — **non-breaking**. When both are set, `secret_key` wins.
- Internal reads, examples, and tests migrated to `secret_key` so the build stays clean under `cargo clippy -- -D warnings`.
- Added a test proving: `secret_key` sets it, `personal_api_key` still works, and `secret_key` wins when both are set (order-independent).

`cargo build`, `cargo test` (workspace), `cargo clippy -- -D warnings`, and `cargo fmt --check` all pass locally.

## Context

Mirrors the same rename in the Python SDK (PostHog/posthog-python#727, draft). Coordinated across the backend SDKs — see the tracking issue PostHog/posthog-js#4046 (rename item only) and the related Slack thread. Naming is aligned so all backend SDKs expose `secret_key` as the canonical name while keeping `personal_api_key` working for backwards compatibility.


turnipdabeets:
**Marked ready for review.**

Naming is settled — this is a generic **`secret_key`** config that accepts **either** a Personal API Key or a Project Secret API Key (both coexist; same `Authorization: Bearer` header). `personal_api_key` stays as a deprecated **config-name** alias that forwards into it — *not* a deprecation of personal keys as a credential. It's a name-level change and is crash-safe with respect to using the key (adds no new error path).

Merge timing still coordinates with the project-secret-key rollout (feature-flagged today) — ready-for-review ≠ merge-now.

api/public-api.txt
```diff
@@ -156,16 +156,17 @@ pub fn posthog_rs::ClientOptionsBuilder::local_evaluation_only(&mut self, bool)
 pub fn posthog_rs::ClientOptionsBuilder::max_batch_size(&mut self, usize) -> &mut Self
 pub fn posthog_rs::ClientOptionsBuilder::max_capture_attempts(&mut self, u32) -> &mut Self
 pub fn posthog_rs::ClientOptionsBuilder::max_queue_size(&mut self, usize) -> &mut Self
-pub fn posthog_rs::ClientOptionsBuilder::personal_api_key<VALUE: core::convert::Into<alloc::string::String>>(&mut self, VALUE) -> &mut Self
```
dustinbyrne, line 159: can we keep this the same?

src/client/async_client.rs
ioannisj: Should we change this warn as well?

src/client/mod.rs
```diff
@@ -398,6 +400,19 @@ impl ClientOptionsBuilder {
...
         Ok(self.build_unchecked()?.sanitize())
     }
+
+    /// Deprecated alias for [`secret_key`](Self::secret_key).
+    ///
+    /// Kept for backwards compatibility. When both are set, `secret_key` wins.
+    #[deprecated(
+        note = "use `secret_key` instead; it accepts a Personal API Key or a Project Secret API Key"
+    )]
+    pub fn personal_api_key(&mut self, value: impl Into<String>) -> &mut Self {
```
ioannisj: Like Dustin's comment above this should probably change to 

```suggestion
    pub fn personal_api_key<VALUE: Into<String>>(&mut self, value: VALUE) -> &mut Self {
```

src/client/mod.rs
```diff
@@ -398,6 +400,19 @@ impl ClientOptionsBuilder {
...
+        if self.secret_key.is_none() {
```
ioannisj: should we remove the guard here so that last builder call always wins? I think that's how the deprecated alias used to work

api/public-api.txt
turnipdabeets: Done — restored `personal_api_key` to its original `<VALUE: Into<String>>` signature, so its public API is unchanged; only `secret_key` is added (41652f0).

src/client/async_client.rs
turnipdabeets: Updated — both the async and blocking local-eval fallback warnings now reference `secret_key` (41652f0).

src/client/mod.rs
turnipdabeets: Applied your suggestion — signature restored to `<VALUE: Into<String>>` (41652f0).

src/client/mod.rs
turnipdabeets: Done — removed the guard so the last builder call wins (transparent forward into `secret_key`), matching the original alias behavior. Updated the test to cover both call orders (41652f0).

ioannisj (approved):

PR #175, opened 2026-07-02
feat: immediate-delivery capture (capture_immediate / capture_batch_immediate)

eli-r-ph:

`capture` / `capture_batch` are fire-and-forget: since v0.14.0 they enqueue onto the background transport worker and never report the outcome. That's the right default for analytics, but it leaves no path for a caller that must know a batch actually persisted before advancing its own durable state — the motivating case being a server-side importer that commits an upstream offset only once the events are safely ingested. Today such consumers are pinned to 0.13.3 (the last release where `capture_batch` returned a `Result`).

This restores an inline confirmed-delivery path as an **additive, opt-in** API without disturbing the fire-and-forget default:

- `Client::capture_immediate(event)` and `Client::capture_batch_immediate(events, historical_migration)` on both the async and blocking clients.
- They send **inline** (bypassing the worker) and return `Result<CaptureSummary, Error>` once the request reaches a terminal outcome, retrying transient failures per the client's existing retry configuration.
- They reuse the existing sans-IO core (`v1_capture` / `v0_capture` / `retry` helpers), so both the v0 and `capture-v1` pipelines are covered with no duplicated protocol logic.

New public `CaptureSummary` (`#[non_exhaustive]`, accessor-only): `submitted()`, `not_persisted()`, `all_persisted()`, and `event_results()` (under `capture-v1`). On v0 a `2xx` persists the whole batch (`all_persisted()` always true); on `capture-v1` the per-event verdicts are surfaced, so a `2xx` with `drop`/`retry` verdicts reports as not fully persisted.

Design notes:

- **Naming**: `*_immediate` mirrors the `captureImmediate` method in posthog-node (send inline, resolve on completion). Unlike Node's `Promise<void>`, the Rust API keeps the richer `Result<CaptureSummary>` so callers can inspect per-event durability.
- Fire-and-forget remains the primary, recommended API and is unchanged.
- The immediate variants **do not** fire `on_error` hooks — the returned `Result` is the delivery signal, avoiding double-reporting. `on_error` stays the observability channel for the fire-and-forget path.
- **De-duplication**: the I/O-free setup (event prep, request building, payload encoding, URL resolution) lives in shared `prepare_immediate` / `build_attempt_parts` helpers in `v0_capture` / `v1_capture`. Each of the four client methods keeps only its transport-specific send loop — async awaits `tokio::time::sleep` between retries, blocking uses `std::thread::sleep` — so there is no new runtime coupling and no protocol logic duplicated across clients.
- Disabled clients and empty / fully `before_send`-filtered batches return a default `CaptureSummary` without sending a request.

`tests/test_capture_immediate.rs` with httpmock coverage across all four permutations (async x blocking, v0 x capture-v1): single/batch success, v1 partial-persist reporting, retryable-status-then-success and inline partial-retry pruning within one call, retry-budget exhaustion returning `Err`, terminal statuses without retry, historical-migration flag on the wire, gzip (query param + `Content-Encoding` header) on v0, disabled/empty no-ops, and an explicit assertion that immediate calls never fire `on_error`.

The retryable-exhaustion and terminal-no-retry cases are parameterized over the full shared status sets (`408/500/502/503/504` retryable; `400/401/402/403`, plus a bare `429` on v0, terminal) in every client/pipeline module, and the retry-then-empty-success case now asserts `not_persisted()` / `all_persisted()` directly.

Verified `cargo test`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `RUSTDOCFLAGS=-D warnings cargo doc` clean across the feature matrix (default async v0, `capture-v1`, `--no-default-features` blocking v0/v1). Existing suites still pass (224 unit + integration). Public API snapshot regenerated via `scripts/check-public-api.sh --update`, and the downstream `capture-v1-rs-smoke` e2e harness still builds against this branch.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.sampo/changesets/confirmed-capture-delivery.md
+++ b/.sampo/changesets/confirmed-capture-delivery.md
@@ -1,9 +1,0 @@
----
-cargo/posthog-rs: minor
----
-
-Confirmed-delivery capture for callers that must know a batch persisted before advancing their own durable state (e.g. a server-side importer committing an upstream offset). `Client::capture_confirmed` and `Client::capture_batch_confirmed` send inline — bypassing the background transport worker — and return `Result<CaptureSummary, Error>` once the request reaches a terminal outcome, retrying transient failures per the client's existing retry configuration. They are available on both the async and blocking clients and on both the v0 and `capture-v1` pipelines, reusing the same sans-IO request-build/retry/response-classification core as the fire-and-forget path.
-
-The new public `CaptureSummary` (`#[non_exhaustive]`, read through accessors) reports the outcome: `submitted()` (events sent after `before_send` filtering), `not_persisted()`, and `all_persisted()`. On the v0 pipeline a `2xx` persists the whole batch, so `all_persisted()` is always true; on `capture-v1` the backend returns per-event verdicts, so a `2xx` can still leave events unpersisted — `not_persisted()` counts submitted events without an `ok`/`warning` verdict, and `event_results()` (under `capture-v1`) exposes the full per-event map.
-
-Fire-and-forget `capture`/`capture_batch` remain the primary, recommended API and are unchanged. The confirmed variants do NOT fire `on_error` hooks: the returned `Result` is the delivery signal, so there is no double-reporting. Disabled clients and an empty (or fully `before_send`-filtered) batch return a default `CaptureSummary` without sending a request. New public export: `CaptureSummary`.
--- a/.sampo/changesets/confirmed-capture-delivery.md
+++ b/.sampo/changesets/confirmed-capture-delivery.md
@@ -1,0 +1,9 @@
+---
+cargo/posthog-rs: minor
+---
+
+Confirmed-delivery capture for callers that must know a batch persisted before advancing their own durable state (e.g. a server-side importer committing an upstream offset). `Client::capture_confirmed` and `Client::capture_batch_confirmed` send inline — bypassing the background transport worker — and return `Result<CaptureSummary, Error>` once the request reaches a terminal outcome, retrying transient failures per the client's existing retry configuration. They are available on both the async and blocking clients and on both the v0 and `capture-v1` pipelines, reusing the same sans-IO request-build/retry/response-classification core as the fire-and-forget path.
+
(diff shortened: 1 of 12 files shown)


marandaneto:
@eli-r-ph python has this https://github.com/PostHog/posthog-python/blob/6f75afe77ff059e4f3b0b6b7b30912612a7b5ff1/posthog/client.py#L284-L285 which bypasses queuing the events, i think something like this could be used in the rust async client so you can await the request

Node has `captureImmediate` methods that bypasses queuing

the other option is an on_success hook very similar to on_error hook (same shape, but rather for success use cases)
i think this is also a nice to have since its been requested before in other SDKs

the other option is to change the hook to have 2 methods, on_success and on_error, since its behind the v1 feature, we could break this API and make it nicer, and ofc rename it to something that makes more sense like idk `http_completion_handler` or something similar

i think all of those approaches are maybe a bit nicer than the one suggested in this PR

WDYT? @PostHog/team-client-libraries opinions?

eli-r-ph:
Thanks for the reply, I'll take a look and update this accordingly - @PostHog/team-client-libraries please do chime in if you have additional thoughts on the matter, want to make sure I'm adding something that is in line with wider needs, doesn't just unblock internal SDK use cases 👍

dustinbyrne:
I like the idea of standardizing on something like `captureImmediate`/`capture_immediate`. I think the global client setting for Python doesn't offer enough control (i.e., cannot easily mix immediate and deferred capture without instantiating an additional client).

`capture_confirmed` is effectively the same as Node's `captureImmediate` though, isn't it? If it's just a matter of standardizing on a single naming scheme, let's go with what's already public.

For languages that have idiomatic return types that indicate a success/failure result, I think it's more expected that we prefer those over communicating via `on_error`/`on_success` hooks.

eli-r-ph:
👋 hey @dustinbyrne made a round of updates based on you an @marandaneto 's callouts, let me know if this feels like the right direction or needs larger rethink 🙇

dustinbyrne (approved):

dustinbyrne:
looks clean, nice addition

PR #177, opened 2026-07-03
fix(capture): clamp Retry-After to retry_max_backoff_ms

eli-r-ph:

Cross-SDK capture retry parity. posthog-rs previously bounded a server `Retry-After` only by an internal 1-day `RETRY_BACKOFF_CAP`, far larger than any sane analytics retry wait. This clamps `Retry-After` to `retry_max_backoff_ms` (default 30s) so a single wait can no longer exceed the caller's configured max backoff, unifying the default 30s retry-wait ceiling with posthog-go and posthog-python.

- `Retry-After` stays a *minimum* (client waits the larger of the configured backoff and the clamped `Retry-After`); the configured backoff is never truncated.
- Removes the internal 1-day cap; an absurd user-set `retry_max_backoff_ms` is the user's own ceiling.
- Applies to the shared v0 + v1 backoff path (`backoff_duration`); feature-flags/poller are unaffected (they pass no `Retry-After`).

- Updated `retry.rs` unit tests: a hostile/large `Retry-After` now clamps to `retry_max_backoff_ms` (was the 1-day cap).
- Updated the three integration retry-after tests to use a max backoff above the 1s header they assert (since `Retry-After` is now clamped to that ceiling).
- `cargo test` and `cargo test --features capture-v1` green; `cargo clippy --features capture-v1` clean.

## :pencil: Checklist

- [x] Behavioral change (`Retry-After` now bounded by `retry_max_backoff_ms`); changeset added.

### If releasing new changes

- [x] Added a sampo changeset (`minor`)

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8) as part of a cross-SDK capture retry unification (paired with posthog-go #255 and the posthog-python capture-v1 stack). Prod telemetry showed the backend sends `Retry-After: 1s` and retryable responses are ~1 in 18M, so the cap only ever guards pathological values; 30s (the existing default max backoff) is the agreed unified ceiling.

### DIFF
--- a/.sampo/changesets/retry-after-clamp.md
+++ b/.sampo/changesets/retry-after-clamp.md
@@ -1,0 +1,5 @@
+---
+cargo/posthog-rs: minor
+---
+
+Clamp a server `Retry-After` to `retry_max_backoff_ms` (default 30s) instead of an internal 1-day cap, so a single retry wait never exceeds the configured max backoff. `Retry-After` still acts as a minimum and the configured backoff itself is never truncated. This unifies the default retry-wait ceiling (30s) with posthog-go and posthog-python.
--- a/.sampo/changesets/retry-after-clamp.md
+++ b/.sampo/changesets/retry-after-clamp.md
@@ -1,5 +1,5 @@
 ---
-cargo/posthog-rs: minor
+cargo/posthog-rs: patch
 ---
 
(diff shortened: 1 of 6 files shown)


.sampo/changesets/retry-after-clamp.md
```diff
@@ -0,0 +1,5 @@
+---
+cargo/posthog-rs: minor
```
marandaneto: i'd say its a patch

marandaneto (approved):

PR #186, opened 2026-08-06
feat: attach trace context to exceptions

hpouillot:

**Why:** Rust exception events need trace and span identifiers so Error Tracking can link failures to distributed traces and logs.

Adds an atomic `trace_context(trace_id, span_id)` builder to `CaptureExceptionOptions`. Automatic correlation is dependency-free: applications configure a client-level `trace_context_provider` callback using whichever telemetry stack and version they already own. Explicit per-exception context takes precedence.

Provider callbacks run synchronously at the capture site, are required to return an atomic pair, and have panics isolated so instrumentation cannot crash the application.

- `cargo test`
- `cargo test --all-features`
- `cargo clippy --all-features --lib -- -D warnings`
- `cargo fmt --check`
- `cargo doc --all-features --no-deps`

## :pencil: Checklist

### If releasing new changes

- [x] Added a Sampo changeset.

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,8 +27,6 @@ os_info = "3.14"
 sha1 = "0.10"
 regex = "1.10"
 tracing = "0.1"
-opentelemetry = { version = "0.32", default-features = false, features = ["trace"], optional = true }
-tracing-opentelemetry = { version = "0.33", default-features = false, optional = true }
 tokio = { version = "1", features = [
     "rt",
     "sync",
@@ -48,7 +46,6 @@ serde_json = "1.0"
 futures = "0.3"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 eyre = "0.6.12"
-opentelemetry_sdk = { version = "0.32", default-features = false, features = ["trace"] }
 
 [features]
 default = ["async-client", "error-tracking"]
@@ -57,7 +54,6 @@ async-client = ["tokio"]
 capture-v1 = ["brotli", "zstd"]
 test-harness = []
 error-tracking = ["dep:backtrace", "dep:findshlibs"]
-opentelemetry = ["error-tracking", "dep:opentelemetry", "dep:tracing-opentelemetry"]
 
 [workspace]
 members = [".", "compliance/adapter"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ async-client = ["tokio"]
 capture-v1 = ["brotli", "zstd"]
 test-harness = []
 error-tracking = ["dep:backtrace", "dep:findshlibs"]
-opentelemetry = ["dep:opentelemetry", "dep:tracing-opentelemetry"]
+opentelemetry = ["error-tracking", "dep:opentelemetry", "dep:tracing-opentelemetry"]
 
 [workspace]
 members = [".", "compliance/adapter"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,8 @@ os_info = "3.14"
 sha1 = "0.10"
 regex = "1.10"
 tracing = "0.1"
+opentelemetry = { version = "0.32", default-features = false, features = ["trace"], optional = true }
+tracing-opentelemetry = { version = "0.33", default-features = false, optional = true }
 tokio = { version = "1", features = [
     "rt",
     "sync",
@@ -46,6 +48,7 @@ serde_json = "1.0"
 futures = "0.3"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 eyre = "0.6.12"
+opentelemetry_sdk = { version = "0.32", default-features = false, features = ["trace"] }
 
 [features]
 default = ["async-client", "error-tracking"]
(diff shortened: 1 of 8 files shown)


src/error_tracking.rs
```diff
@@ -379,10 +404,25 @@ where
...
         fingerprint,
         level,
     } = options;
 
+    #[cfg(feature = "opentelemetry")]
+    let current_context = (trace_id.is_none() || span_id.is_none())
+        .then(current_trace_context)
+        .flatten();
+    #[cfg(feature = "opentelemetry")]
+    let (trace_id, span_id) = match current_context {
```
marandaneto: blocking: Trace context needs to remain atomic. If a caller supplies only one ID here, the other is taken from the ambient span and can belong to a different trace; the added test currently demonstrates that mismatched pair. Please only use ambient IDs when neither explicit ID is supplied, or expose/validate a paired trace-context option.

src/error_tracking.rs
```diff
@@ -402,12 +442,34 @@ where
...
     exception.write_into(&mut event, et_options)?;
     Ok(event)
 }
 
+#[cfg(feature = "opentelemetry")]
+fn current_trace_context() -> Option<(String, String)> {
+    use opentelemetry::trace::TraceContextExt;
+    use tracing_opentelemetry::OpenTelemetrySpanExt;
+
+    let context = tracing::Span::current().context();
```
marandaneto: blocking: This only reads context from a compatible `tracing-opentelemetry` layer, so native OpenTelemetry contexts attached through `opentelemetry::Context::attach` or `FutureExt::with_context` silently produce no IDs despite the API promising the current OpenTelemetry span. Please support `opentelemetry::Context::current()` as a fallback when the current tracing span has no valid context.

Cargo.toml
```diff
@@ -54,6 +57,7 @@ async-client = ["tokio"]
 capture-v1 = ["brotli", "zstd"]
 test-harness = []
 error-tracking = ["dep:backtrace", "dep:findshlibs"]
+opentelemetry = ["dep:opentelemetry", "dep:tracing-opentelemetry"]
```
marandaneto: blocking: With `default-features = false, features = ["opentelemetry"]`, this enables the dependencies but leaves the entire exception API disabled behind `error-tracking`, making the advertised feature ineffective by itself. Please make this feature include `error-tracking` and add a check for that feature combination.

marandaneto (commented):
Three blocking findings around trace-context correctness and feature behavior.

Cargo.toml
marandaneto: also worth adding this feat flag to the ci matrix

hpouillot: Fixed in 08de402. The opentelemetry feature now includes error-tracking, and CI tests --no-default-features --features opentelemetry.

src/error_tracking.rs
```diff
@@ -379,10 +404,25 @@ where
...
+    let current_context = (trace_id.is_none() || span_id.is_none())
+        .then(current_trace_context)
+        .flatten();
+    #[cfg(feature = "opentelemetry")]
+    let (trace_id, span_id) = match current_context {
+        Some((current_trace_id, current_span_id)) => (
+            trace_id.or(Some(current_trace_id)),
+            span_id.or(Some(current_span_id)),
+        ),
+        None => (trace_id, span_id),
```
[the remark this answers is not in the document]
hpouillot: Fixed in 08de402. Explicit trace context is now accepted only as an atomic trace/span pair, and ambient context is used only when no explicit pair is supplied.

src/error_tracking.rs
hpouillot: Fixed in 08de402. Context resolution now tries the current tracing-opentelemetry span first and falls back to opentelemetry::Context::current(). Added a test using Context::attach().

src/error_tracking.rs
hpouillot: Fixed in 08de402. The separate builders were replaced with trace_context(trace_id, span_id), so identifiers from explicit and ambient sources can no longer be mixed. Added explicit-pair precedence coverage.

marandaneto (approved):

hpouillot:
Reworked automatic correlation in `2c6ba9d` to avoid coupling the SDK to OpenTelemetry versions:

- removed the OpenTelemetry dependencies and feature
- added a client-level `trace_context_provider` callback
- kept trace/span IDs atomic and explicit per-exception context authoritative
- isolated provider panics
- added provider, precedence, and panic tests

Applications can now adapt native OpenTelemetry, `tracing-opentelemetry`, or another tracing system using dependencies they already own.

src/error_tracking.rs
```diff
@@ -2779,10 +2890,72 @@ mod tests {
...
             "checkout-error"
         );
         assert_eq!(json["properties"]["$exception_level"], "warning");
     }
+
+    #[test]
+    fn build_exception_event_reads_trace_context_provider() {
+        let options = ErrorTrackingOptionsBuilder::default()
+            .trace_context_provider(|| {
+                Some(TraceContext::new("provider-trace-id", "provider-span-id"))
```
cat-ph, line 2909: do we need to enforce/validate those to be valid hex/length (I think 32 hex for trace ids, 16 hex for span)? 🤔 I think we do _some form of it_ in python: https://github.com/PostHog/posthog-python/blob/c72f8d94f5e95ccd2deac0c5ff4ea6cc15b9ed02/posthog/exception_utils.py#L108-L114

cat-ph (approved):

marandaneto (approved):

PR #200, opened 2026-08-07
fix(flags): bound cohort recursion in local evaluation

slshults:
## Problem

Local flag evaluation can abort the calling process.

Cohort resolution recurses through nested cohort references (`match_cohort_by_id` -> `match_property_group` -> `match_property_group_values` -> `match_nested_cohort` -> back to `match_cohort_by_id`) with nothing bounding the recursion. A cohort that references itself, directly or through another cohort, recurses until the thread stack is exhausted. Reproduced on `main` before the fix:

```
thread '...test_self_referencing_cohort_is_inconclusive' has overflowed its stack
fatal runtime error: stack overflow, aborting
process didn't exit successfully (signal: 6, SIGABRT)
```

A Rust stack overflow is not a catchable panic. It aborts the whole process, not just the evaluating thread, so a single bad definitions response takes down the customer's server, and takes it down again on restart as soon as the poller refetches. That is an outage, not a blip.

Greptile flagged this on #199, where it showed up only because that branch was stacked on #187 before the squash merge. It is not #199's code, so it gets its own PR.

## Why a cycle check alone is not enough

A cycle guard bounds repeats, not depth. A chain of distinct cohorts, where cohort N references cohort N+1, is perfectly acyclic and still recurses once per link. It reaches the identical abort.

That case also slips past the one accidental protection that exists today. A single deeply nested property group is rejected at parse time by `serde_json`'s recursion limit of 128, so the whole `LocalEvaluationResponse` fails to deserialize and the previous good cache is kept. But a chain spreads its depth across many separate shallow map entries, none of which comes close to that limit, so it deserializes cleanly and only blows up later during evaluation.

So the fix bounds both.

## What changed

A `HashSet` of the cohort IDs on the **active resolution path** is threaded through the four private resolution functions, giving two bounds for one mechanism:

- an ID already on the path is a cycle
- the path length is capped at `MAX_COHORT_RESOLUTION_DEPTH` (100, against real nesting of a handful of levels)

Both return `CohortMatchError::InvalidDefinition`, whose `requires_server_evaluation()` is already `true`, so the error short-circuits the enclosing group and propagates. The flag ends up inconclusive and falls back to `/flags`. It cannot silently resolve to `false`, which would be a wrong answer rather than a safe fallback. That was traced explicitly through `match_property_group_values`, where a plain `Inconclusive` would instead be deferred and could be masked by a decisive sibling in an OR group.

IDs are removed as the recursion unwinds, so this tracks the active path rather than everything ever visited. The same cohort referenced twice down sibling branches, which is a diamond and not a cycle, still resolves normally.

No public API change (`api/public-api.txt` regenerates clean). Only private function signatures moved.

## Tests

- self-referencing cohort is inconclusive
- two-cohort cycle is inconclusive, and the error names the cycle
- chain deeper than the limit is inconclusive
- chain within the limit still resolves, so the bound does not break realistic nesting
- the same cohort referenced down sibling branches still resolves, which is the guard against implementing this as a naive "visited forever" set
- end to end through `LocalEvaluator::evaluate_flag` with a cyclic manifest, so the regression is pinned at the level a real bad payload would hit

Written test-first: the self-reference test aborted the test binary with SIGABRT before the fix.

## Note on how the depth case was found

The cycle fix was written first, and a review pass specifically asked whether the fix was complete. It was not: the acyclic-chain bypass was caught and reproduced at a 5000-link chain. Worth recording, because "fixed the stack overflow" would have been half true with only the cycle guard, and the remaining half is just as fatal.

---


src/feature_flags.rs
```diff
@@ -926,11 +927,29 @@ impl CohortMatchError {
...
 }
 
+/// How deep a chain of cohort references local evaluation will follow.
+///
+/// A cycle check alone bounds repeats, not depth: a chain of distinct cohorts is
+/// acyclic and still recurses once per link. Each cohort is a separate shallow
+/// entry in the manifest, so such a chain also clears `serde_json`'s own nesting
+/// limit on the way in, leaving nothing else between a long chain and an
+/// exhausted stack. Real cohort nesting is a handful of levels at most.
+const MAX_COHORT_RESOLUTION_DEPTH: usize = 100;
```
marandaneto, line 937: blocking: Recursion limit can still overflow the stack

A 100-cohort path creates several Rust stack frames per link, so the process can abort before this guard runs on smaller stacks. This still reproduces at this HEAD: running the deep-chain test binary with `RUST_MIN_STACK=262144` aborts with `stack overflow`. Please make traversal iterative, or use a substantially lower limit validated against realistic thread-stack budgets.

marandaneto:
pushed a fix

marandaneto (approved):

PR #206, opened 2026-08-10
feat!: make EvaluateFlagsOptions non-exhaustive

marandaneto:

`EvaluateFlagsOptions` currently allows external struct literals. Adding a field after v1 would break those callers.

This marks the struct as `#[non_exhaustive]` before the stable release. Future options can then be added in minor releases. Callers construct the options with `EvaluateFlagsOptions::default()` and assign the fields they need.

The compliance adapter and integration tests now use that construction pattern. The public API snapshot and rustdoc describe the new contract.

This prepares the options type for future context work discussed in #151.

- `cargo fmt --check`
- `cargo test --test test_evaluate_flags --test test_local_evaluation`
- `cargo test --no-default-features --test test_evaluate_flags`
- `cargo check --workspace --all-targets`
- `cargo doc --no-deps`

## :pencil: Checklist

### If releasing new changes


marandaneto:
do not merge because its a breaking change, i want to stack a few changes together before releasing v1
you can review the changes tho

ioannisj (approved):

Issue #207, opened 2026-08-10
Prepare and release posthog-rs| posthog v1.0


marandaneto:
## Goal

Release the first stable major version of the Rust SDK after the V1 capture pipeline is fully rolled out. The release should make the supported capture path and public API clear, remove migration-only APIs, and provide an upgrade guide for existing `0.x` users.

## V1 capture rollout

- [ ] Complete the production rollout of the `capture-v1` pipeline.
- [x] Make V1 capture the default supported capture path.
- [x] Remove the V0 capture plumbing and its Cargo/build feature before releasing v1.
- [x] Run the V1 SDK compliance suite for async and blocking clients and confirm all supported compression and retry behavior.
- [x] Confirm capture, immediate capture, error tracking, feature flag events, historical migration, `before_send`, and `on_error` all use the V1 path correctly.
- [x] Document the capture behavior changes and any migration steps from V0.

## AI event capture readiness

- [x] Confirm whether posthog-rs must expose a dedicated `capture_ai()` API before v1.
- [x] If required, implement `capture_ai()` using the V1 pipeline with an isolated queue and retry buffer.
- [x] Ensure AI events do not inherit standard V1 capture payload limits that could reject large AI payloads.
- [ ] Add tests covering large AI payloads, retries, batching, and isolation from the standard capture queue.
- [ ] Document `capture_ai()` and the recommended way to send AI events.
- [x] If `capture_ai()` is deferred, get explicit approval to release with AI events using `capture()` and document the resulting payload limitations.

## Public API cleanup

- [x] Merge #208 to remove the deprecated feature flag methods and the deprecated `personal_api_key` builder alias, and rename `LocalEvaluationConfig::personal_api_key` to `secret_key`.
- [x] Audit public option and response types for changes that should be made before the API becomes stable.
- [x] Merge #206. This is work in progress and makes `EvaluateFlagsOptions` non-exhaustive before v1.
- [x] Resolve the breaking TLS feature configuration proposed in #201.

## Package and documentation

- [ ] Complete #178 as part of the v1 release plan and decide how long `posthog-rs` and `posthog` releases will be mirrored.
- [x] Update the Rust SDK docs and examples so they only use the v1 APIs and `secret_key` terminology.
- [x] Publish a `0.x` to `1.0` migration guide covering removed APIs, Cargo features, capture behavior, and any crate rename.
- [ ] Confirm the public API snapshot contains only APIs intended to be supported throughout the v1 release line.

## Release validation

[...]

PR #209, opened 2026-08-10
feat!: use capture v1 by default

marandaneto:

The V1 capture pipeline is ready to become the supported default as part of #207. Existing applications should still be able to remain on V0 while they prepare to migrate.

This adds `capture-v1` to the default Cargo features and documents how to opt out by disabling default features. The CI matrix now tests the V1 default and an asynchronous V0 fallback. The compliance adapters disable dependency defaults so the V0 and V1 suites continue to build the intended pipeline.

- `cargo fmt -- --check`
- `cargo build --verbose`
- `cargo build --verbose --no-default-features --features async-client,error-tracking`
- `cargo test`
- `cargo test --workspace`
- `cargo test --no-default-features --features async-client,error-tracking`
- `cargo test --no-default-features --features capture-v1`
- `cargo test --no-default-features --features error-tracking`
- `cargo test --no-default-features --features error-tracking,capture-v1`
- `cargo test --verbose --features e2e-test --no-default-features`
- `cargo clippy -- -D warnings`
- `cargo clippy --no-default-features --features async-client,error-tracking -- -D warnings`
- `cargo publish --dry-run --locked`
- Built the V0 and V1 compliance adapters in release mode with their Dockerfile Cargo commands and verified that only the V1 adapter enabled `capture-v1`.

## :pencil: Checklist

- [ ] I added tests to verify the changes.

### If releasing new changes


README.md
```diff
@@ -11,6 +11,19 @@ SDK usage examples and code snippets live in the official documentation so they
...
+The V1 capture pipeline is enabled by default. To continue using the legacy V0
+pipeline, disable the default features and explicitly enable the other features
+your application needs:
+
+```toml
+posthog-rs = { version = "1", default-features = false, features = ["async-client", "error-tracking"] }
+```
+
+This example retains the default asynchronous client and error tracking while
+opting out of `capture-v1`.
```
marandaneto: TBD: if we remove the capture-v0 and v1 altogether and strip all capture v0 code as well

marandaneto:
do not merge because its a breaking change, i want to stack a few changes together before releasing v1
this PR will likely change its impl so not worth reviewing yet, we need to figure [this](https://github.com/PostHog/posthog-rs/pull/209#discussion_r3752511094) out

README.md
dustinbyrne: i think we should drop it. imo it's additional surface area that we need to maintain for little/no value

dustinbyrne (approved):

PR #215, opened 2026-08-14
fix: normalize SDK timestamps to UTC

marandaneto:

V0 capture payloads serialize event timestamps as naive date-time strings without a UTC offset. This makes the wire timestamp ambiguous even though `Event::set_timestamp` already converts the supplied value to the equivalent UTC instant.

This change stores the V0 wire timestamp as `DateTime<Utc>`, so both single-event and batch capture payloads serialize timestamps as RFC 3339 UTC values with a `Z` suffix. For example, `2023-01-01T10:00:00.123+03:00` is sent as `2023-01-01T07:00:00.123Z`. V1 serialization remains unchanged and continues to send the same UTC instant.

The regression coverage verifies that behavior directly. `v0_serializes_non_utc_timestamp_as_equivalent_utc_instant` checks V0 conversion from `+03:00` to `Z`. `v1_event_preserves_utc_timestamp_serialization` checks the same instant on V1. The existing `event_timestamp_is_capture_time_not_publish_time` transport test now accepts only RFC 3339 timestamps, so its default and `capture-v1` runs also prove both wire formats include a UTC offset while preserving capture-time stamping.

- `cargo fmt --check`
- `cargo test --lib` - 224 tests passed.
- `cargo test --lib --features capture-v1` - 262 tests passed.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.sampo/changesets/regal-stormcaller-tuulikki.md
+++ b/.sampo/changesets/regal-stormcaller-tuulikki.md
@@ -1,0 +1,5 @@
+---
+cargo/posthog-rs: patch
+---
+
+Normalize SDK timestamps to UTC
--- a/src/client/transport.rs
+++ b/src/client/transport.rs
@@ -1461,17 +1461,11 @@ mod tests {
         );
     }
 
-    /// Parse either an RFC3339 string (v1 timestamps / v0 `sent_at`) or a naive
-    /// datetime (v0 event timestamps serialize without an offset) as UTC.
+    /// Parse an RFC3339 wire timestamp and normalize it to UTC.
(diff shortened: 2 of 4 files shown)


arnohillen (approved):
<small>(created by claude code)</small>

[arno's agent] Verified at 4c540a40: set_timestamp stores naive_utc() from tz-aware input so the .and_utc() relabel in InnerEvent::from_event is invariantly safe, all v0 sends funnel through it, and the +03:00-to-07:00Z tests would fail on a relabel; 46/46 and 111/111 compliance green.

Issue #150, opened 2026-06-19
Add group_identify helper


marandaneto:
## Summary

Add a first-class `group_identify` helper to the Rust SDK for creating/updating group properties.

## Motivation

The Rust SDK supports group analytics for events through `Event::add_group`, but it doesn't expose a helper for the `$groupidentify` event. Other server-side SDKs, such as Python, support this directly:

```python
posthog.group_identify('company', 'company_id_in_your_db', {
    'name': 'Awesome Inc.',
    'employees': 11,
})
```

Without a Rust helper, users must manually construct the `$groupidentify` event and know the required property names.

## Proposed API

Something like:

```rust
client.group_identify(
    "company",
    "company_id_in_your_db",
    properties,
).await?;
```

For the blocking client:

```rust
client.group_identify("company", "company_id_in_your_db", properties)?;
```

This should send a `$groupidentify` event with the expected properties:

```json
{
  "event": "$groupidentify",
  "properties": {
    "$group_type": "company",
    "$group_key": "company_id_in_your_db",
    "$group_set": {
      "name": "Awesome Inc.",
      "employees": 11
    }
  }
}
```

## Acceptance criteria

- Async and blocking clients expose a `group_identify` helper.
- The helper accepts serializable group properties.
- Tests cover the generated `$groupidentify` event payload.
- Docs can use the helper in group analytics examples.

PR #216, opened 2026-08-14
feat: add alias helper for merging distinct IDs

### LINKED ISSUE #149: Add alias helper
## Summary

Add a first-class `alias` helper to the Rust SDK so users don't need to manually construct a `$create_alias` event.

## Motivation

Other server-side SDKs expose an alias helper for assigning multiple distinct IDs to the same person. For example, Python supports:

```python
posthog.alias(previous_id='frontend_id', distinct_id='backend_id')
```

Rust can currently send the underlying event manually, but there is no ergonomic API. This makes the Rust SDK look like it lacks part of user identification compared to SDKs such as Python, Node, Go, PHP, and Ruby.

## Proposed API

Something like:

```rust
client.alias("frontend_id", "backend_id").await?;
```

For the blocking client:

```rust
client.alias("frontend_id", "backend_id")?;
```

This should send a `$create_alias` event equivalent to the documented API shape:

```json
{
  "event": "$create_alias",
  "distinct_id": "frontend_id",
  "properties": {
    "alias": "backend_id"
  }
}
```

## Acceptance criteria

- Async and blocking clients expose an alias helper.
- The helper sends the correct `$create_alias` payload.
- Tests cover the generated event payload.
- Docs can point users to the helper instead of showing manual event construction.

HughScott2002:

closes #149. (Left unlinked rather than auto-closing — the issue's last acceptance
criterion is about the docs on posthog.com pointing at the helper, which isn't
this repo. Close it whenever you think it's covered.)

Adds a first-class `alias` helper so users don't hand-build the `$create_alias` event:

```rust
client.alias("anon-abc123", "user-42");
```

Available on both the async and blocking clients with the same signature.

### Two places I deviated from the issue — please sanity-check these

**1. The event carries `distinct_id` inside `properties`, which the issue's JSON snippet omits.**

The issue specifies:

```json
{
  "event": "$create_alias",
  "distinct_id": "frontend_id",
  "properties": { "alias": "backend_id" }
}
```

Every other PostHog SDK also writes `distinct_id` into `properties`:

| SDK                                                                                                                  | `properties` keys      |
| -------------------------------------------------------------------------------------------------------------------- | ---------------------- |
| [posthog-python](https://github.com/PostHog/posthog-python/blob/master/posthog/client.py)                            | `distinct_id`, `alias` |
| [posthog-js-lite](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-core/src/index.ts) (`aliasStateless`) | `distinct_id`, `alias` |
| [posthog-php](https://github.com/PostHog/posthog-php/blob/master/lib/Client.php)                                     | `distinct_id`, `alias` |

I followed the SDKs rather than the snippet, so Rust doesn't emit an event missing a field the other libraries all send. Happy to switch if the snippet is authoritative and the others are carrying legacy baggage.

**2. The signature is sync and infallible, not `async` + `Result`.**

The issue proposes `client.alias(a, b).await?`. This crate's `capture()` is sync and returns nothing even on the async client — fire-and-forget is the core model, and delivery failures surface through the `on_error` hook. Building an alias event cannot fail (two strings into a map), so a `Result` would always be `Ok` and every caller would have to `?` or discard it, and there is nothing to `.await`.

Following `capture()` also means the async and blocking call sites are character-for-character identical, so docs and snippets are portable between them. Easy to change if you'd rather match the other SDKs' shape.

### Other behavior worth flagging

- **Blank IDs are dropped, not sent.** A blank ID on either side can't describe a merge. posthog-python drops with a warning; this does the same, rejecting `""` and whitespace-only. IDs that pass are sent exactly as given, never trimmed.
- **Person profile processing is left on**, as a merge requires — the helper doesn't stamp the `$process_person_profile: false` that `Event::new_anon` sets.
- **Construction lives in `Event::alias`**, so both clients share one definition of the payload and the guard, and neither can drift.
- **Not added:** a global `posthog_rs::alias()` alongside `global::capture`. #149 asks only for the two clients, and #150/#152 raise the same question — probably better decided once across all three helpers. Say the word and I'll add it here.

`tests/test_alias.rs` — 6 tests, mirrored across the blocking and async client modules. Verified against all five relevant feature combinations (30 test executions):

| Combination                                                   | Result   |
| ------------------------------------------------------------- | -------- |
| default (async + v0)                                          | 6 passed |
| `--features capture-v1` (async + v1)                          | 6 passed |
| `--no-default-features` (blocking + v0)                       | 6 passed |
| `--no-default-features --features error-tracking`             | 6 passed |
| `--no-default-features --features capture-v1` (blocking + v1) | 6 passed |

Coverage: payload shape, `$process_person_profile` not disabled, blank-ID drop on each side, `Into<String>` ergonomics with mixed `&str`/`String`, disabled-client no-op, and distinct event UUIDs across repeated calls.

The emitted event is identical on both capture pipelines — nothing in it is lifted into the V1 `options` object — so one file with a `CAPTURE_PATH` switch covers V0 and V1 rather than the split-file treatment error tracking needs.

I captured the real request body off a socket to confirm the wire format on both pipelines:

```json
{
  "event": "$create_alias",
  "distinct_id": "anon-abc123",
  "properties": {
    "alias": "user-42",
    "distinct_id": "anon-abc123",
    "$is_server": true
  }
}
```

**Mutation testing.** To check the tests actually catch regressions rather than just passing, I wrote 11 deliberately broken versions of `Event::alias` and confirmed each one fails the suite: swapped IDs, dropped `properties.distinct_id`, dropped `properties.alias`, wrong event name, event attributed to the new ID, guard removed, guard ignoring whitespace, guard checking only one side, merge marked personless, and a reused event UUID. All 11 are caught.

The UUID case was a genuine hole on the first pass — nothing asserted uniqueness, so `Uuid::nil()` passed everything. Since PostHog deduplicates on event UUID, that would have silently dropped every alias after the first. `repeated_aliases_get_distinct_event_uuids` closes it.

**Not covered, for transparency:** no run against a live PostHog instance (the `e2e-test` feature needs `POSTHOG_RS_E2E_TEST_API_KEY`, which CI holds), so this verifies the emitted bytes match the other SDKs rather than that the merge lands server-side. Unicode/very long IDs, concurrent `alias` calls, and `before_send` interaction are also untested.

Also run locally and passing: `cargo fmt -- --check`, `cargo clippy -- -D warnings`, `cargo test --doc`, `cargo doc --no-deps` (no rustdoc warnings), `cargo publish --dry-run --locked`, and `scripts/check-public-api.sh`.

`api/public-api.txt` is regenerated via `scripts/check-public-api.sh --update`; the diff is one added line.

Two unrelated tests fail on my machine, on `main` as well as on this branch: `error_tracking::tests::from_error_builds_exception_list_with_stacktrace` and `inlined_frames_emit_expanded_groups`, both with `expected hex instruction_addr, got ""`. That's local stack symbolication on NixOS, not anything this branch touches — verified against a clean worktree at `main`.

## :pencil: Checklist

### If releasing new changes

<sub>`sampo` isn't published as a binary I could install locally, so `.sampo/changesets/add-alias-helper.md` is hand-written to the format in RELEASING.md (`cargo/posthog-rs: minor`, matching the changesets from #184 and #186). Happy to regenerate it if the file needs anything else.</sub>

---

<sub>Written with AI assistance (Claude Code). I reviewed every line, ran the full feature matrix locally, and the SDK cross-references and mutation results above are ones I verified rather than took on trust. This is my first contribution here so please let me know if anything doesn't match how you like PRs shaped.</sub>

### DIFF
--- a/.sampo/changesets/add-alias-helper.md
+++ b/.sampo/changesets/add-alias-helper.md
@@ -1,0 +1,5 @@
+---
+cargo/posthog-rs: minor
+---
+
+Add a `Client::alias` helper that merges two distinct IDs onto the same person, so callers no longer need to hand-build the `$create_alias` event and know that `distinct_id` is written both at the top level and inside `properties`. Available on the async and blocking clients with the same fire-and-forget signature as `capture`. A blank ID on either side cannot describe a merge, so the event is dropped with a warning rather than sent.
--- a/api/public-api.txt
+++ b/api/public-api.txt
@@ -125,6 +125,7 @@ pub fn posthog_rs::CaptureSummary::not_persisted(&self) -> usize
 pub fn posthog_rs::CaptureSummary::submitted(&self) -> usize
 pub struct posthog_rs::Client
 impl posthog_rs::Client
+pub fn posthog_rs::Client::alias<P: core::convert::Into<alloc::string::String>, D: core::convert::Into<alloc::string::String>>(&self, P, D)
 pub fn posthog_rs::Client::capture(&self, posthog_rs::Event)
 pub fn posthog_rs::Client::capture_batch(&self, alloc::vec::Vec<posthog_rs::Event>, bool)
 pub async fn posthog_rs::Client::capture_batch_immediate(&self, alloc::vec::Vec<posthog_rs::Event>, bool) -> core::result::Result<posthog_rs::CaptureSummary, posthog_rs::Error>
--- a/src/client/async_client.rs
+++ b/src/client/async_client.rs
@@ -198,6 +198,38 @@ impl Client {
         }
     }
 
+    /// Merge two distinct IDs onto the same person by sending a `$create_alias`
+    /// event.
+    ///
+    /// See <https://posthog.com/docs/product-analytics/identify#alias-assigning-multiple-distinct-ids-to-the-same-user>.
+    ///
+    /// # Parameters
+    ///
+    /// - `previous_id`: ID already known to PostHog, such as an anonymous ID.
+    /// - `distinct_id`: ID it should be merged into, such as a logged-in user ID.
+    ///
+    /// # Remarks
+    ///
+    /// Fire-and-forget, like [`Client::capture`]. A blank ID on either side
+    /// cannot describe a merge, so the event is dropped with a warning rather
+    /// than sent.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # async fn example() {
+    /// let client = posthog_rs::client("phc_project_api_key").await;
+    ///
+    /// // The visitor browsed anonymously, then logged in.
+    /// client.alias("anon-abc123", "user-42");
+    /// # }
+    /// ```
+    pub fn alias<P: Into<String>, D: Into<String>>(&self, previous_id: P, distinct_id: D) {
+        if let Some(event) = Event::alias(previous_id.into(), distinct_id.into()) {
+            self.capture(event);
+        }
+    }
+
     /// Flush queued events, returning once the worker has attempted delivery of
     /// everything queued before this call. Transient failures are kept for retry
     /// (the call still returns without error). A no-op for disabled clients.
--- a/src/client/blocking.rs
+++ b/src/client/blocking.rs
@@ -202,6 +202,36 @@ impl Client {
         }
     }
 
+    /// Merge two distinct IDs onto the same person by sending a `$create_alias`
+    /// event.
+    ///
+    /// See <https://posthog.com/docs/product-analytics/identify#alias-assigning-multiple-distinct-ids-to-the-same-user>.
+    ///
+    /// # Parameters
+    ///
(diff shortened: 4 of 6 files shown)


marandaneto:
thanks @HughScott2002

marandaneto (approved):

marandaneto:
> Commits must have verified signatures.

@HughScott2002 can you sign your commits and rewrite the commit history? otherwise we cant merge it

HughScott2002:
Sorry about that, I didn't have commit signing set up when I opened this. It's sorted now.

Force-pushed as `3c5d397` with an SSH signature and GitHub is showing it as verified. The tree is identical to before, so the signature and the SHA are the only things that changed.

Heads up that the new SHA seems to have re-triggered the first-time contributor gate on the workflows, so they'll need approving again whenever you get a chance.

Thanks for your patience and review.

PR #219, opened 2026-08-23
fix: avoid self-joining the transport worker

marandaneto:

Dropping a client or calling async `shutdown()` from a transport callback can run teardown on the transport worker itself. Waiting for shutdown completion or joining that same thread deadlocks.

## Changes

Centralize blocking transport teardown and guard async shutdown with the same worker-thread check. Worker-thread callers queue shutdown without waiting or joining, while external callers keep the existing durable close and join behavior.

Focused callback tests cover both `Drop` and async `shutdown()`, repeated teardown, and concurrent close behavior. The full library test suite passes with 257 tests. Blocking teardown was also checked with `--no-default-features --features error-tracking`.

Autoreview completed for commit `613aba1` against `origin/v1` with no actionable findings.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/src/client/async_client.rs
+++ b/src/client/async_client.rs
@@ -918,22 +918,60 @@ impl Drop for Client {
     /// `shutdown` can't run in a destructor), so dropping a `Client` inside an
     /// async task blocks that executor thread until the drain completes (up to
     /// `shutdown_timeout_ms` plus any in-flight request) — prefer an explicit
-    /// `shutdown().await` first, which makes this a no-op.
+    /// `shutdown().await` first, which makes this a no-op. A drop from a transport
+    /// callback instead queues shutdown without waiting for or joining the current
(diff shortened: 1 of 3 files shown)


ioannisj (commented):
Do we need a similar treatment for the async shutdown() version?

ioannisj (approved):

marandaneto:
Good question. The async `shutdown()` path had the same worker-thread deadlock risk. Fixed in signed commit `613aba1`: it now queues shutdown but skips awaiting completion and joining when called from the transport worker. I also added a callback regression test. The focused tests and all 257 library tests pass.

marandaneto:
Superseded by #233, which applies the same signed fixes and regression tests to a clean branch based on `main`. Retargeting this branch directly would include unrelated `v1` history.

PR #230, opened 2026-08-23
refactor: reduce duplicated SDK evaluation and capture logic

marandaneto:

Several production policies and test fixtures are duplicated across async and blocking clients, capture delivery, and feature evaluation. This integration PR keeps the remaining focused refactors together so their combined behavior can be validated before they land individually.

## Changes

Combine the remaining focused component branches into one `v1` integration branch. The diff now contains shared retry classification, capture preprocessing, feature-flag event hosting and evaluation state, capture batch routing, and client/capture test consolidation. The independently merged feature evaluation, debug-image filtering, and teardown work are now part of the updated `v1` base.

`cargo fmt --check`, default tests, no-default-feature tests, blocking error-tracking tests, all-feature tests, public API snapshot validation, and independent threading, feature semantics, and simplicity reviews.

Autoreview completed for this exact branch head and base with no actionable findings.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/src/client/common.rs
+++ b/src/client/common.rs
@@ -48,7 +48,7 @@ pub(super) fn flag_event_dedup_cache() -> FlagEventDedupCache {
     Mutex::new(HashMap::new())
 }
 
-pub(super) fn apply_capture_defaults(event: &mut Event, defaults: &CaptureDefaults) {
+fn apply_capture_defaults(event: &mut Event, defaults: &CaptureDefaults) {
     if defaults.disable_geoip {
         event.insert_prop_default("$geoip_disable", serde_json::Value::Bool(true));
     }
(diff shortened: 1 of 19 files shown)


src/client/common.rs
```diff
@@ -272,7 +320,155 @@ pub(super) fn extract_flag_details(response: FeatureFlagsResponse) -> DetailedFl
...
+/// retain their concrete HTTP and async/blocking boundaries; this state owns
+/// only the runtime-neutral decisions on whether to fetch and how to merge.
+pub(super) struct EvaluationState {
+    distinct_id: String,
+    options: EvaluateFlagsOptions,
+    records: HashMap<String, EvaluatedFlagRecord>,
+    request_id: Option<String>,
+    errors_while_computing: bool,
+    quota_limited: bool,
+}
```
dustinbyrne, line 333: hmm i just saw this in another PR?

dustinbyrne (approved):

src/client/common.rs
marandaneto: Yes — this block is also in #224. #230 is the umbrella integration PR that intentionally combines all of the focused duplication-refactor branches so the complete tree can be validated together; #224 is the focused PR for this specific `EvaluationState` change. This is the same implementation carried through the stack, not a second independent copy.

PR #233, opened 2026-08-24
fix: avoid self-joining the transport worker

marandaneto:

Dropping a client or calling async `shutdown()` from a transport callback can run teardown on the transport worker itself. Waiting for shutdown completion or joining that same thread deadlocks.

This replaces #219 with a branch based on `main`, where the affected transport code is already released. The fix can then flow into `v1` through the normal `main` merge.

## Changes

Centralize blocking transport teardown and guard async shutdown with the same worker-thread check. Worker-thread callers queue shutdown without waiting or joining, while external callers keep the existing durable close and join behavior.

- Applied both commits cleanly to `origin/main` with no conflicts.
- `cargo fmt --check`
- `cargo test --lib`: 231 passed
- `cargo test --no-default-features --features error-tracking --lib`: 229 passed
- Autoreview against `origin/main`: no actionable findings

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.sampo/changesets/fabled-firebringer-lemminkainen.md
+++ b/.sampo/changesets/fabled-firebringer-lemminkainen.md
@@ -1,0 +1,5 @@
+---
+cargo/posthog-rs: patch
+---
+
+Prevent transport worker deadlocks during client shutdown.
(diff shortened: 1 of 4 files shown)


marandaneto:
@ioannisj since you approved already https://github.com/PostHog/posthog-rs/pull/219 but target wrong branch

ioannisj (approved):

PR #238, opened 2026-08-27
fix(flags): align local string matching with the flags service

### LINKED ISSUE #78019: Case folding and numeric stringification are inconsistent between SDK local evaluation and the flags service
While shipping `starts_with`/`ends_with` local evaluation across the SDKs (#72992), a pattern emerged: our case-insensitive property matching is inconsistent between the SDKs and the flags service, and within the flags service itself. Filing this so we can settle it once instead of per code review.

## How the flags service folds case

In `rust/feature-flags/src/properties/property_matching.rs`:

- `icontains` / `starts_with` / `ends_with` (and negations): `to_ascii_lowercase()`, ASCII-only. The code comment says it's a deliberate perf choice.
- `exact` / `is_not`: full Unicode `to_lowercase()`.

So the server already applies two different folding rules depending on operator.

## What the SDKs use for the icontains/starts_with/ends_with family

| SDK | Folding | Matches server? |
|---|---|---|
| PHP | `strtolower` (ASCII-only) | ✅ |
| Ruby | `downcase(:ascii)` as of PostHog/posthog-ruby#229 | ✅ |
| Elixir | ASCII-only in PostHog/posthog-elixir#192 (pending) | ✅ |
| Python | `casefold()` (most aggressive Unicode; `"ss"` matches `"ß"`) | ❌ |
| Go | `strings.ToLower` (Unicode) | ❌ |
| Node | `toLowerCase()` (Unicode) | ❌ |
| Rust SDK | `to_lowercase()` (Unicode) | ❌ |
| .NET | `OrdinalIgnoreCase` (Unicode simple folding) | ❌ |

Concretely: person property `"Äbc"` with `starts_with: "ä"` matches locally in Python/Go/Node/Rust/.NET and does **not** match on `/flags`. Same flag, same person, different answer depending on evaluation path — exactly the divergence local evaluation is supposed to avoid.

## Adjacent inconsistencies in the same code

- The Rust SDK's `exact` operator uses `eq_ignore_ascii_case` while the server's `exact` is Unicode — the mirror image of the divergence above.
- Numeric stringification: the server renders a float-typed `323.0` as `"323.0"` (serde `Display`), as do Ruby and Python, but Go and Node render `"323"`, so a JSON `323.0` property matches `ends_with: "3"` locally but not remotely. .NET additionally used the host locale for decimal separators (`3.14` → `"3,14"` on de-DE), fixed in PostHog/posthog-dotnet#270.

## Options

1. **Standardize the SDKs on ASCII-only folding** for the icontains/starts/ends family to match the server. PHP and Ruby are already there; the rest are one-line changes plus tests. The numeric stringification fixes for Go/Node belong in the same sweep.
2. **Make the server Unicode-aware** and leave the SDKs alone. Friendlier for international text, but changes existing server behavior, and the ASCII choice was deliberate.
3. **Leave it and document** the edge case.

I lean toward 1, plus an explicit decision on whether the server's Unicode `exact`/`is_not` is intentional. Non-ASCII values differing only by case are rare, but today the answer depends on which SDK and which operator you hit, and that's the worst version.

marandaneto:

Local feature flag evaluation must match the currently released `/flags` service until its behavior changes. The service applies boolean-like truthiness before ordinary array membership for `exact` and `is_not`. This includes all-boolean arrays and the vacuous truthiness of an empty array.

The separate backend proposal in [PostHog/posthog#90694](https://github.com/PostHog/posthog/pull/90694) would replace those semantics. This PR intentionally does not adopt that draft behavior.

This fixes PostHog/posthog#78019 by:

- applying the service's boolean-like filter precedence before ordinary array membership
- preserving `is_not` as the direct complement of `exact`
- using ordinary ANY membership only for arrays that are not entirely boolean-like
- keeping serde_json stringification and Rust Unicode lowercase for `exact` and `is_not`
- keeping ASCII-only lowercase for contains, prefix, and suffix operators
- adding final-sigma, dotted-I, boolean-array, and empty-array regression vectors

The patch changeset for `cargo/posthog-rs` describes the complete matching update.

- Red: `cargo test test_exact_boolean_coercion_matches_flags_service -- --nocapture` failed at `exact` with filter `false` and property `"banana"`.
- Green: the same focused test passed after porting the service truthiness helpers.
- `cargo test test_property_case_folding_matches_flags_service -- --nocapture`
- `cargo test`
- `cargo fmt --all -- --check`
- `cargo clippy -- -D warnings`
- Autoreview with `autoreview --mode local`

The test suite emits existing deprecation warnings from httpmock-based tests. No test failed.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi using the address-pr-comments, check-pr, autoreview, and karpathy-guidelines skills. Verification showed that Rust and serde_json already match the service's Unicode and JSON behavior, so production changes remain limited to boolean-like precedence.

### DIFF
--- a/src/feature_flags.rs
+++ b/src/feature_flags.rs
@@ -1717,29 +1717,38 @@ mod tests {
             match_property(&property, &HashMap::from([("key".to_string(), actual)])).unwrap()
         };
 
-        // Exact matching stringifies both sides and applies Unicode lowercase.
-        assert!(matches("exact", json!("PRO"), json!("pro")));
-        assert!(matches("exact", json!("Ä"), json!("ä")));
-        assert!(!matches("exact", json!("ß"), json!("ss")));
-        assert!(!matches("exact", json!("Σ"), json!("ς")));
-        assert!(matches("exact", json!(323), json!("323")));
-        assert!(matches("exact", json!(["FREE", "PRÖ"]), json!("prö")));
-        assert!(!matches("is_not", json!(["FREE", "PRÖ"]), json!("prö")));
-        assert!(matches("is_not", json!(["FREE", "PRÖ"]), json!("team")));
-
         // Serde preserves a float's decimal point; exact matching must not collapse it to an integer.
         assert_eq!(value_to_string(&json!(323.0)), "323.0");
-        assert!(matches("exact", json!("323.0"), json!(323.0)));
-        assert!(!matches("exact", json!("323"), json!(323.0)));
 
-        // Substring and anchored operators fold ASCII only.
-        assert!(matches("icontains", json!("ADMIN"), json!("admin-user")));
-        assert!(!matches("icontains", json!("Ä"), json!("äbc")));
-        assert!(matches("not_icontains", json!("Ä"), json!("äbc")));
-        assert!(!matches("starts_with", json!("Ä"), json!("äbc")));
-        assert!(matches("not_starts_with", json!("Ä"), json!("äbc")));
-        assert!(!matches("ends_with", json!("Ä"), json!("bcä")));
-        assert!(matches("not_ends_with", json!("Ä"), json!("bcä")));
+        let cases = [
+            // Exact matching stringifies both sides and applies Unicode lowercase.
+            ("exact", json!("PRO"), json!("pro"), true),
+            ("exact", json!("Ä"), json!("ä"), true),
+            ("exact", json!("ß"), json!("ss"), false),
+            ("exact", json!("Σ"), json!("ς"), false),
+            ("exact", json!(323), json!("323"), true),
+            ("exact", json!(["FREE", "PRÖ"]), json!("prö"), true),
+            ("is_not", json!(["FREE", "PRÖ"]), json!("prö"), false),
+            ("is_not", json!(["FREE", "PRÖ"]), json!("team"), true),
+            ("exact", json!("323.0"), json!(323.0), true),
+            ("exact", json!("323"), json!(323.0), false),
+            // Substring and anchored operators fold ASCII only.
+            ("icontains", json!("ADMIN"), json!("admin-user"), true),
+            ("icontains", json!("Ä"), json!("äbc"), false),
+            ("not_icontains", json!("Ä"), json!("äbc"), true),
+            ("starts_with", json!("Ä"), json!("äbc"), false),
+            ("not_starts_with", json!("Ä"), json!("äbc"), true),
+            ("ends_with", json!("Ä"), json!("bcä"), false),
+            ("not_ends_with", json!("Ä"), json!("bcä"), true),
+        ];
+
+        for (operator, expected, actual, should_match) in cases {
+            let result = matches(operator, expected.clone(), actual.clone());
+            assert_eq!(
+                result, should_match,
+                "operator {operator} comparing {actual} against {expected}"
+            );
+        }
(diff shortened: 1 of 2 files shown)


src/feature_flags.rs
```diff
@@ -1707,6 +1705,43 @@ mod tests {
...
+        assert!(!matches("exact", json!("323"), json!(323.0)));
+
+        // Substring and anchored operators fold ASCII only.
+        assert!(matches("icontains", json!("ADMIN"), json!("admin-user")));
+        assert!(!matches("icontains", json!("Ä"), json!("äbc")));
+        assert!(matches("not_icontains", json!("Ä"), json!("äbc")));
+        assert!(!matches("starts_with", json!("Ä"), json!("äbc")));
+        assert!(matches("not_starts_with", json!("Ä"), json!("äbc")));
+        assert!(!matches("ends_with", json!("Ä"), json!("bcä")));
+        assert!(matches("not_ends_with", json!("Ä"), json!("bcä")));
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 74efa03 — the comparison matrix is now table-driven, with each failure reporting the operator and compared values.

src/feature_flags.rs
```diff
@@ -1577,13 +1581,7 @@ fn match_property(
...
 
 fn compare_values(a: &serde_json::Value, b: &serde_json::Value) -> bool {
-    // Case-insensitive string comparison
-    if let (Some(a_str), Some(b_str)) = (a.as_str(), b.as_str()) {
-        return a_str.eq_ignore_ascii_case(b_str);
-    }
-
-    // Direct comparison for other types
-    a == b
+    value_to_string(a).to_lowercase() == value_to_string(b).to_lowercase()
```
ioannisj: [suggestion] Worth hoisting the person-side conversion out of the loop here? `compare_values` now stringifies and lowercases both sides on every call, and the `exact`/`is_not` arms call it once per element of `property.value` (line 1434), so an allow-list of N values does 4N allocations where the old `eq_ignore_ascii_case` path did none.

The flags service already does this, `rust/feature-flags/src/properties/property_matching.rs:220` pulls `let target = to_string_representation(override_value).to_lowercase()` out before iterating. Same semantics, N allocations instead of 4N.

src/feature_flags.rs
```diff
@@ -1459,40 +1459,44 @@ fn match_property(
...
+                .to_ascii_lowercase()
+                .contains(&search_str.to_ascii_lowercase())
         }
         "starts_with" => {
             let prop_str = value_to_string(value);
             let search_str = value_to_string(&property.value);
             prop_str
-                .to_lowercase()
-                .starts_with(&search_str.to_lowercase())
+                .to_ascii_lowercase()
```
ioannisj: [nit] Do we have a reference for folding these two ASCII-only? The flags service `OperatorType` (`rust/feature-flags/src/properties/property_models.rs:11`) has no `StartsWith`/`EndsWith` at all, and posthog-node doesn't either, so the changeset's "same case folding as the flags service" really only holds for `icontains`.

The narrowing is real, `ends_with: "Ä"` used to match `"bcä"` and doesn't anymore. I think consistency with `icontains` is the right call, but could the changeset say that rather than service parity?

ioannisj (approved):

PR #239, opened 2026-08-27
feat: report $release_id from POSTHOG_RELEASE_ID

ablaszkiewicz:
## Problem

A compiled Rust binary carries no release. A JavaScript build injects `$release_id` into its bundle, and a mobile app reads its version from the OS, but a Rust binary has neither.

## Change

Report a `$release_id` on `$exception` events, from either an explicit `release_id` client option or the `POSTHOG_RELEASE_ID` environment variable. `posthog-cli release resolve` creates the release and prints its id; that id reaches the SDK one of two ways:

- **Build time (baked in).** Set the `release_id` option — typically `option_env!("POSTHOG_RELEASE_ID")` — so the id is compiled into the binary. A shipped binary then self-identifies with nothing set at runtime. This survives code signing (the id is in the linked binary before you sign) and needs no deploy-time configuration.
- **Runtime (env fallback).** Leave the option unset and the SDK reads `POSTHOG_RELEASE_ID` from the environment at launch, so a deploy can supply the release without a rebuild.

**Precedence:** an explicit option wins over the environment.

```rust
let options = ClientOptionsBuilder::default()
    .api_key(key)
    // Baked at build time; falls back to POSTHOG_RELEASE_ID at runtime when unset.
    .release_id(option_env!("POSTHOG_RELEASE_ID").unwrap_or_default())
    .build()?;
```

The server resolves each exception's release by a direct id lookup, so no release name or version has to match anything the app reports. Only `$exception` events carry the property — that is the only event the server resolves a release from. The value is resolved once per capture; an unset or blank value changes nothing; the property is set before `before_send`, so a hook can still drop it. **No binary patching and no code signing.**

**Trade-off.** Baking a different id per release changes the binary, so its debug id changes and each release gets its own symbol set (the runtime-env path keeps one binary and one symbol set across releases). Both are supported; pick per deployment — known-at-build-time favours baking, assigned-after-build favours the env var.

**Contract:** the release id is the UUID that `posthog-cli release resolve` prints. The CLI side is PostHog/posthog#89834: `symbol-sets upload --release-mode=event` uploads the symbols release-independent, and `release resolve` (already on `master`) names the release and prints its id.

This is a second, lighter alternative to the marker-injection approach in #237. The two are independent branches off `main`; pick one.

## How did you test this code?

**Automated** (`cargo test`): unit tests for the value normalization (unset → none, blank → none, whitespace trimmed), the source precedence (explicit option beats the environment; environment used when the option is unset; none when neither is set), the event-name gate (a `$release_id` lands on an `$exception`, not on a `$pageview`, and not at all when unset), and that an explicit `release_id` option reaches the resolved capture defaults. `cargo fmt --check` and `cargo clippy` clean; the public-API snapshot is updated for the new setter.

**End to end against a local PostHog stack**, run by the agent (Claude). The `rust-release-env` example — a plain posthog-rs app with no release code of its own — is built against this branch. `./run` resolves the release, **builds with `POSTHOG_RELEASE_ID` set so `option_env!` bakes the id into the binary**, uploads the symbols release-independent, and runs the binary **with the runtime env unset** — so the reported id can only have come from the build. It is shipped twice (1.0.0, 2.0.0) to show the release tracks the build.

<details>
<summary>posthog-cli — release resolve, then a release-independent <code>symbol-sets upload</code> of the baked build</summary>

```
# the release is named here; its id is baked into the binary at build time
$ posthog-cli release resolve --release-name rust-release-env --release-version 2.0.0
01a043c7-7d1a-0000-3834-390958992b49

# symbols, release-independent — no release name/version here (PostHog/posthog#89834)
# --include-source bundles the .rs files so frames show source context
$ posthog-cli symbol-sets upload --directory target/release --include-source --release-mode=event
INFO  Collected 1251 source files
INFO  Processing dSYM .../rust-release-env.dSYM (UUIDs: 3A474928-6A67-3E44-887A-5CBC62E95BAF)
INFO  --release-mode=event: uploading symbol sets release-independent; the release is carried on each exception as $release_id (POSTHOG_RELEASE_ID)
INFO  Upload summary: 1 chunk(s) uploaded, 0 skipped
```

</details>

<details>
<summary>the app, run with <code>POSTHOG_RELEASE_ID</code> <em>unset</em> — the id is already compiled in</summary>

```
$ unset POSTHOG_RELEASE_ID
$ ./target/release/rust-release-env
starting — release 01a043c7-7d1a-0000-3834-390958992b49 (baked into the binary at build time via option_env!)
stopping
```

</details>

Read back server-side (a dev-login session over the local API): one issue holds both builds, each `$exception` (`$lib = posthog-rs`) carries its baked `$release_id`, cymbal resolved it into a `$exception_release`, and the in-app frames symbolicate to source off the uploaded symbol set:

```
$lib               : posthog-rs
frames (in app)    : crash_site three.rs:5  two two.rs:6  one one.rs:6  main main.rs:41  — symbolicated, with source
1.0.0  $release_id 01a04245-8c54-0000-7530-28eed93002b0 → $exception_release rust-release-env@1.0.0
2.0.0  $release_id 01a043c7-7d1a-0000-3834-390958992b49 → $exception_release rust-release-env@2.0.0
```

| Symbolicated stack trace, with Rust source context | Release resolved via the baked `$release_id` |
| --- | --- |
| stack-trace | release-resolved |

The binary was run with nothing in its environment — the release id it reports was compiled in at build time. The same code shipped as 2.0.0 is a different baked id, resolving to a different release.

## 🤖 Agent context

Authored by Claude in Claude Code, directed by @ablaszkiewicz (DRI). This is the SDK half of a second, env-variable-based approach to native release tracking, offered alongside the marker-injection pair (#237 + PostHog/posthog#89117) for comparison. The `release_id` client option (build-time baking, env var as fallback) was added in response to review feedback on this PR. The end-to-end verification above was run by the agent against a local PostHog dev stack (ingestion + cymbal), with screenshots uploaded via `hogli pr:upload-image`. The example data is invented (`rust-release-env` demo binary) and draws on no customer material.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/release_env.rs
```diff
@@ -0,0 +1,50 @@
...
+/// The environment variable the release id is read from.
+const RELEASE_ID_ENV: &str = "POSTHOG_RELEASE_ID";
+
+/// The release id from `POSTHOG_RELEASE_ID`, read once. `None` when the variable is unset or blank.
+pub(crate) fn release_id() -> Option<&'static str> {
+    static CACHE: OnceLock<Option<String>> = OnceLock::new();
+    CACHE
+        .get_or_init(|| normalize(std::env::var(RELEASE_ID_ENV).ok()))
+        .as_deref()
+}
```
dustinbyrne, line 24: should this be a proc macro? if we want to bake the env var into the binary at build time, this won't work

src/release_env.rs
```diff
@@ -0,0 +1,50 @@
...
+//! This is the deploy-time counterpart to injecting `$release_id` into a web bundle. A native build
+//! has no bundle to inject, so the CLI's `release resolve` prints the created release's id, and the
+//! app is launched with that id in `POSTHOG_RELEASE_ID`. The SDK reads it here and stamps it on
+//! every event, so the server resolves the exception's release by a direct id lookup — no release
+//! name or version has to match anything the app reports.
+
+use std::sync::OnceLock;
+
+/// The environment variable the release id is read from.
+const RELEASE_ID_ENV: &str = "POSTHOG_RELEASE_ID";
```
dustinbyrne, line 16: just a thought - we could fall back to attempting to derive a release id from the git HEAD (`git rev-parse --verify 'HEAD^{commit}'`)

expecting cases where `git` is not found or we're not running in a `.git` repo

marandaneto:
i think we should expose the apps metadata as config, so users would know what to set eg app name, version, build and then we infer the release id in the backend, or even a human readable release identifier
its also ok to provide a release_id env. var. but its less user friendly

eg release = 'myapp@v1.0.0+1'
or app_name = myapp
version = v1.0.0
build =1
or release_git_sha = 'git sha'

this PR only exposes release_id which is a UUID that requires the posthog-cli in advance
just a suggestion, not a blocker

ablaszkiewicz:
Going to close this as I decided there is no point in migrating rust to the new release system. It always builds one chunk. Migration makes setup more complicated and gives no benefit

PR #250, opened 2026-09-08
feat: add capture_ai() on an isolated AI capture lane

eli-r-ph:
Adds `capture_ai()` so Rust callers can send LLM analytics events to the live `POST /i/v1/ai/events` endpoint. AI events ride their own isolated transport lane, so a multi-MB AI event can never delay or displace analytics events. Targets `v1`; all public API changes are additive.

- Adds `capture_ai`, `capture_ai_batch`, `capture_ai_immediate`, `capture_ai_batch_immediate`, and the global `posthog_rs::capture_ai`, mirroring the analytics entry points.
- Adds `Endpoint::CaptureAi` and `CaptureFailure::endpoint()` so an `on_error` hook can tell which lane failed.
- Adds `capture_ai_compression` (mirrors `capture_compression`; unset sends AI bodies uncompressed, zstd recommended) and `capture_ai_max_queue_size` (default 1000).
- The AI lane has its own worker, queue, and retry state, started on the first `capture_ai*` call, and drained by `flush`/`shutdown`/`Drop` alongside the analytics lane.
- The AI lane batches by serialized bytes (5 MiB soft target) and drops events whose `properties` exceed the backend's 8 MiB ceiling locally, logging only the event name and size.
  - The immediate variants send oversize events instead and return the backend's `ai_event_too_big` verdict.
- Neither `capture` nor `capture_ai` inspects the event name; the backend decides routing and reports per event, and that verdict already reaches `on_error`.
  - Deliberate: `$ai_`-prefixed custom names are legitimate analytics traffic, and the analytics lane will start refusing AI names once PostHog/posthog#95466 lands.
- A 2xx with per-event drops and no `on_error` hook now logs one aggregate warning per batch on both lanes; it was silent before.
- The analytics lane is unchanged: the worker and pipeline are parameterized by a lane config whose analytics variant reproduces the previous behavior, and the existing tests pass unmodified.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


marandaneto (commented):
Advisory review: two reproduced behavioral defects. Existing library and AI integration tests pass for both client flavors.

src/client/transport.rs
```diff
@@ -507,12 +750,14 @@ fn run_worker(
                 // `len` is not decremented here: the in-flight counter spans the
                 // whole worker lifecycle (channel + buffer + retries) and is
                 // decremented by the pipeline once a batch is delivered or dropped.
-                if buffer.is_empty() {
-                    buffer_since = Some(clock.now());
-                }
-                buffer.push(*event);
-                if buffer.len() >= flush_at {
-                    send_buffer(&mut pipeline, &mut buffer, max_batch_size, None);
+                if admit_live(
```
marandaneto: blocking: Preserve the carryover event’s flush timer — When admitting an AI event flushes the preceding batch but leaves the new event buffered, `admit_live` starts its timer and returns `true`; this caller then clears `buffer_since` unconditionally. Three roughly 2 MiB events therefore send the first two but leave the third without an automatic interval flush, until another batching threshold or explicit flush/shutdown. Reproduction: reproduced — `review_carryover_flushes_on_interval` advanced a manual clock past the interval and observed one request instead of two; preserving the carryover timer made the test pass.

src/client/blocking.rs
```diff
@@ -289,9 +376,8 @@ impl Client {
     /// calls are no-ops. After shutdown, `capture` drops events. A no-op for
     /// disabled clients.
     pub fn shutdown(&self) {
-        if let Some(transport) = &self.transport {
-            transport.close_blocking();
-        }
+        let lanes: Vec<&TransportHandle> = self.lanes().map(Arc::as_ref).collect();
```
marandaneto: blocking: Close admission before snapshotting shutdown lanes — The post-init recheck still misses an ordering different from the previously resolved thread: shutdown snapshots only analytics, then a first `capture_ai` initializes its lane and completes both open-state checks before analytics is marked closed. Shutdown subsequently closes only its snapshot, returning without draining or joining the AI worker. The async implementation has the same ordering at `src/client/async_client.rs:383–387`. Reproduction: reproduced — `review_ai_init_between_shutdown_snapshot_and_close` paused public blocking shutdown after its snapshot using a test-only barrier, initialized AI capture, and observed the AI lane still open after shutdown returned; marking analytics closed before the snapshot made the test pass.

.sampo/changesets/capture-ai-lane.md
```diff
@@ -0,0 +1,15 @@
+---
+cargo/posthog-rs: minor
```
marandaneto: we can remove htis since its a branch feature

marandaneto (approved):

PR #251, opened 2026-09-09
fix: omit null object properties when serializing events

marandaneto:

Captured custom properties currently send null object members. This implements the serialization contract in https://github.com/PostHog/sdk-specs/pull/60 without narrowing the public property API.

Both v0 and v1 wire serializers use one private helper to omit null object members recursively after hooks and enrichment. Null array entries keep their positions. Empty objects and arrays, other scalar values, and caller-owned events stay unchanged. Queued, batch, immediate, AI-named and exception events use the same boundary.

Generated missing/error feature flag response nulls retain their existing meaning only on `$feature_flag_called`, for `$feature_flag_response` and the exact `$feature/<key>` selected by `$feature_flag`. Privacy filtering remains authoritative. Envelope fields and unrelated caches are unchanged.

Includes a manually authored Sampo patch entry using the repository's required package key.

- Both permanent loopback HTTP tests pass across blocking/async clients, v0/v1 capture, and error tracking enabled/disabled (eight feature combinations).
- Focused serializer, event, envelope and minimal flag privacy unit tests pass in four error-tracking-enabled builds (21 tests per v0 build and 40 per v1 build).
- Formatting and the public API snapshot check pass. Strict Clippy passes for both async builds.
- **Known baseline limitation:** strict blocking Clippy fails on existing `clippy::type_complexity` at `src/client/blocking.rs:706`, also reproduced on unchanged baseline. Both blocking builds pass with only that lint allowed on the command line. No source suppression was added.
- Isolated committed-head review reports no actionable findings. All 25 hosted checks passed for `eb246af5dab6041f1a7fe56cb4db8325b3fd6384`, including the CI test matrix, public API, CodeQL and both SDK compliance workflows. Existing compliance coverage does not prove all null-policy scenarios.
- Full suites, minimum Rust versions, performance, local-poller and retry/compression-specific null behavior were not validated locally. The SDK has no disk-backed event queue to exercise. This draft is not release or universal conformance clearance.

## :pencil: Checklist

### If releasing new changes

The equivalent Sampo patch file was authored manually as permitted by `RELEASING.md`. No release command was run.

### DIFF
--- a/.sampo/changesets/omit-null-event-properties.md
+++ b/.sampo/changesets/omit-null-event-properties.md
@@ -1,0 +1,5 @@
+---
+cargo/posthog-rs: patch
+---
+
+Omit null object properties recursively when serializing captured events, while preserving null array entries and generated feature flag response metadata.
--- a/src/event.rs
+++ b/src/event.rs
@@ -470,15 +470,36 @@ pub struct BatchRequest {
 
 // With `capture-v1` enabled nothing outside tests builds the V0 wire format.
 #[cfg_attr(feature = "capture-v1", allow(dead_code))]
-#[derive(Serialize)]
 pub struct InnerEvent {
-    #[serde(skip_serializing_if = "Option::is_none")]
     api_key: Option<String>,
     uuid: Uuid,
     event: String,
     distinct_id: String,
     properties: HashMap<String, serde_json::Value>,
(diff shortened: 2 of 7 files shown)


marandaneto:
gonna be done server side anyway

PR #257, opened 2026-09-15
feat(client): allow supplying custom reqwest clients

### LINKED ISSUE #256: Allow supplying a custom reqwest client to the SDK
### Use case

I'm embedding the Rust SDK into an application that already owns its HTTP/TLS setup, and there is currently no way to make the SDK share it.

Concretely, our application builds its `reqwest` clients with `rustls-no-provider` plus a `CryptoProvider` (ring) that we install once at process start. We also configure proxies, timeouts, and a shared connection pool in one place. The SDK builds its own `reqwest` clients internally — as far as I can tell, up to three of them (the async/blocking client used for capture and `/flags`, the blocking client in the background capture transport, and the one in the local-evaluation poller) — and none of them can be configured beyond `request_timeout_seconds`.

That means:

- The SDK's clients don't pick up our process-installed crypto provider / TLS configuration, which is the part that actually blocks us.
- Our proxy and connection-pool configuration doesn't apply to SDK traffic.
- We end up with extra connection pools alongside the one the application already maintains.

### What I checked first

`before_send` doesn't help here — it hooks event content, not transport. `request_timeout_seconds` is the only transport knob today and it only covers timeouts.

### Proposed API shape

Two new optional `ClientOptions` fields (builder setters), defaulting to `None` so existing behaviour is unchanged:

```rust
let options = ClientOptionsBuilder::default()
    .api_key(token)
    .http_client(my_reqwest_client)                   // reqwest::Client
    .blocking_http_client(my_blocking_reqwest_client) // reqwest::blocking::Client
    .build()?;
```

Semantics I'd suggest: when a client is supplied it is used exactly as given (the caller owns timeouts/TLS/proxy/pool), so `request_timeout_seconds` would not be applied to it; when absent, behaviour is byte-for-byte what it is today.

Two options rather than one because the blocking capture transport needs a `reqwest::blocking::Client` even on the async client, so a single field couldn't cover both.

I'm happy to send a PR, and equally happy to adjust the shape (naming, one option vs. two, or a client-factory/trait instead of concrete reqwest types) based on what you'd prefer. Per CONTRIBUTING.md I wanted to agree the shape here first.

amitksingh1490:
Closes #256 (API shape proposed there — happy to change it, see below).

## Motivation

I'm embedding this SDK in an application that already owns its HTTP and TLS configuration, and there's currently no way to make the SDK share it.

Our application builds its `reqwest` clients with `rustls-no-provider` plus a `ring` `CryptoProvider` that we install once at process start, and it configures proxies, timeouts and a shared connection pool in one place. The SDK builds up to three `reqwest` clients internally — the async/blocking client used for capture and `/flags`, the blocking client in the background capture transport, and the one in the local-evaluation poller — and none of them can be configured beyond `request_timeout_seconds`. So the SDK's clients don't pick up our crypto provider or TLS setup (the part that actually blocks us), our proxy/pool configuration doesn't apply to SDK traffic, and we carry extra connection pools alongside the application's.

`before_send` doesn't cover this — it hooks event content, not transport.

### 1. Custom reqwest clients

Two new optional `ClientOptions` builder setters, both defaulting to `None`:

```rust
let options = ClientOptionsBuilder::default()
    .api_key(token)
    .http_client(my_client)                   // reqwest::Client
    .blocking_http_client(my_blocking_client) // reqwest::blocking::Client
    .build()?;
```

- `http_client` is used by the async client constructor and the `AsyncFlagPoller`.
- `blocking_http_client` is used by the blocking client constructor, both capture-transport `Pipeline::new` impls (v0 and v1), and the blocking `FlagPoller`.

Semantics: **a supplied client is used exactly as given** — the caller owns timeouts, TLS, proxies and pooling — so `request_timeout_seconds` is deliberately *not* applied to it. This is documented on both fields. When the options are unset, behaviour is unchanged.

Two options rather than one because the capture transport needs a `reqwest::blocking::Client` even under the async client, so a single field couldn't cover both paths.

Threading notes:
- Both options live on `ClientOptions` only. `LocalEvaluationConfig` is unchanged, because its fields are public and adding one would break existing struct literals; the pollers instead get a `pub(crate) set_http_client`, mirroring the existing `set_on_error`.
- The blocking `FlagPoller::start` built a *second* client inside its polling thread, which would have ignored a supplied client. It now clones the one the poller already holds, so the polling loop uses the caller's client too. This also removes a duplicate client construction on the existing path.
- `ClientOptions` still derives `Clone` (both reqwest clients are `Clone`) and still doesn't need `Debug`.

### 2. `background_transport` option

Same motivation, second half of the problem. Constructing a client unconditionally spawns the background batching worker — a dedicated std thread which itself builds a *blocking* reqwest client (spinning up another runtime thread) — even for callers who only ever use `capture_immediate`/`capture_batch_immediate`. `Client::drop` then does a blocking worker join when `shutdown().await` wasn't called, which is painful when dropping a client inside an async task.

For an embedder running on its own async runtime, and using only immediate capture, that's two threads and a blocking drop for machinery it never touches.

```rust
let options = ClientOptionsBuilder::default()
    .api_key(token)
    .background_transport(false) // default: true
    .build()?;
```

With it off, no worker thread and no blocking HTTP client are created. `capture_immediate`/`capture_batch_immediate` and feature-flag calls work normally (they use the client's own HTTP client), and `shutdown`/`Drop` return without blocking because there's nothing to join. Default is `true`, so existing behaviour is untouched.

Fire-and-forget `capture`/`capture_batch`/`alias`/`group_identify` have no worker to enqueue onto, so they drop their events and **log a single warning** (via a `Once`) pointing at the immediate methods. I picked warn-once over silent-drop because silently discarding events a caller explicitly sent seems worth one line of output; it stays silent for disabled clients, where dropping is already the documented behaviour. Happy to switch it to a silent drop if you'd prefer.

One thing worth flagging from auditing the `transport.is_some()` assumptions: the **panic hook** keyed off `is_disabled()` alone. Without a worker it would have latched the single process-wide panic hook and then dropped every `$exception` it built, so `should_capture_global_panics`/`capture_panic`/`install_panic_hook` now also require a background transport. The other call sites (`flush`, `shutdown`, `Drop`, `pending_events`, the flag-called event host, `flush_blocking_timeout`, `on_transport_worker`) already handled `None` correctly.

## Tests

`tests/test_custom_http_client.rs` sets a distinctive default header on a supplied client and asserts it arrives at an `httpmock` server:

- async: `capture_immediate` through a supplied `reqwest::Client`, plus a control test showing the default client carries no such header when the option is absent;
- blocking: capture + `flush()` through the background transport with a supplied `blocking_http_client`.

`tests/test_background_transport.rs` covers the new option: immediate capture still reaching the server through the supplied client with no worker; fire-and-forget capture (plus `capture_batch`/`alias`) reaching the server zero times even after `flush().await`; `shutdown().await` and `drop` both returning well under a second; a default-is-on control; and the blocking equivalent asserting no transport is constructed.

I checked the behavioural tests are not vacuous — reverting each option to be ignored makes them fail. The teardown test deliberately points at a mock that stalls every response for 10s, so a client that *did* spawn a worker blows the time budget rather than passing incidentally (a refused connection would fail fast and prove nothing).

Verified across the default, `capture-v1` and `--no-default-features` feature combinations; all green. `cargo fmt -- --check` and `cargo clippy -- -D warnings` (the CI command) pass. `cargo clippy --all-targets -- -D warnings` and `cargo clippy --no-default-features -- -D warnings` fail on `main` too, before this change — deprecated `httpmock::assert_hits` in existing tests and two pre-existing lib lints; this branch adds no new warnings of its own. I skipped the `e2e-test` suite as it needs network credentials.

`api/public-api.txt` was regenerated with `scripts/check-public-api.sh --update` (cargo-public-api 0.52.0 on `nightly-2026-06-12`); the diff is exactly the three new builder setters.

## Note on public API

Per CONTRIBUTING.md this adds public API, so I opened #256 first with the use case and the proposed shape, and I've left this as a **draft** pending a maintainer's view on that issue. I'm very happy to adjust — naming, one option instead of two, whether `background_transport` belongs here at all or is better expressed some other way, or an entirely different shape such as a client-factory/trait rather than concrete `reqwest` types if you'd rather not widen the public `reqwest` surface. Treating all of that as an open question, not a decision.


marandaneto:
see https://github.com/PostHog/posthog-rs/issues/256#issuecomment-5793462898
@amitksingh1490 thanks for the PR
closing as one part is already addressed
and for https://github.com/PostHog/posthog-rs/issues/262 i suggest opening a new PR so we dont mix up the 2 things

